PR #6, opened 2020-04-28
Fold `get_export` into `exports`

### LINKED ISSUE #4: Python idioms in the API
The README mentions that:

> So far this extension has been written by folks who are primarily Rust programmers, so it's highly likely that there's some faux pas in terms of Python idioms. Feel free to send a PR to help make things more idiomatic if you see something!

Indeed, after using the API for a while I noticed a few aspects that aren't quite in line with Python conventions, though some of them are harder to ignore than others.

  1. Nullary getter methods like `ExportType.name` that always succeed and always return the same value are a pretty clear-cut case for using Python properties instead. This is, by far, the least idiomatic aspect of the library, and it's very easy to miss a `()` that feels like it shouldn't be there.
  2. There are lots of unary setter methods like `Config.debug_info` named after the property they're changing, and there is a good case for using Python setter-only properties instead. (As a bonus, if wasi-c-api ever adds getters for them, they could be made readable in a backwards compatible way.) This doesn't change much for e.g. `Config`, but for `Linker` for example `allow_shadowing` works like a simple setter with no side effects, but `define_wasi`, while similarly imperatively named and used, irreversibly changes its state.
  3. Nullary casting methods like `ExternType.func_type` feel very odd overall.
    1. They're named like accessors, but they work like `dynamic_cast`.
    2. The naming is inconsistent for conversion in different directions—`FuncType`→`ExternType` conversion is done by `as_extern`, but `ExternType`→`FuncType` conversion is performed by `func_type`.
    3. My impression so far is that most uses of these casting methods are in contexts where a specific type is expected, in which case it would be more convenient if they raised `TypeError` instead of returning `None`. (If they did, another method would be necessary to dispatch on an unknown type.)
  4. `Val.get_*` introduces a third naming scheme for type conversion functions, one that clashes in meaning with e.g. `Instance.get_export`.
  5. `*.get_export()` methods seem like they would only fail in truly exceptional conditions and so also appear good candidates for raising an exception instead of returning `None`.

To summarize:
  1. I would change simple getters and setters to be properties.
  2. I would rename all casting methods to `as_*()`.
  3. I would change `*.get_export()` to raise if the export is not found (this is particularly important for `Caller.get_export()`, since currently it looks like there is no way to distinguish the cases of "export does not exist" and "the caller is gone" at all, even for debugging, and the exception message would allow that).
  4. I would *probably* change the dynamic casting methods to raise if the type is wrong, and add a matching `is_*` method for the cases where dynamic dispatch is indeed necessary. I'm not completely certain about this one, but it seems that currently, dynamic dispatch would look something like:
   ```python
   if extern.memory():
       do_something_with_memory(extern.memory())
   elif extern.table():
       do_something_with_table(extern.table())
   ```
   so changing it to:

   ```python
   if extern.is_memory():
       do_something_with_memory(extern.as_memory())
   elif extern.is_table():
       do_something_with_table(extern.as_table())
   ```

   doesn't really make it any worse.

What do you think? All of these changes are likely to break most existing code, so I'd rather discuss them before working on a PR.

Also, I should mention that "do nothing" is an option. In general Python libraries are a lot looser with conventions than libraries in many other languages; I'd rather see the API improved, personally, but the current one isn't drastically more inconsistent than what one would expect in general.

alexcrichton:
Remove `Instance.get_export` as well as `Caller.get_export`. The former
is replaced with string indexing on `Instance.exports` and the latter is
replaced with string indexing on `Caller`

Closes #4

### DIFF
--- a/wasmtime/_func.py
+++ b/wasmtime/_func.py
@@ -141,7 +141,22 @@ class Caller(object):
     def __init__(self, ptr):
         self.__ptr__ = ptr
 
-    def get_export(self, name):
+    def __getitem__(self, name):
+        """
+        Looks up an export with `name` on the calling module.
+
+        If `name` isn't defined on the calling module, or if the caller has gone
+        away for some reason, then this will raise a `KeyError`. For more
+        information about when this could fail see the `get` method which
+        returns `None` on failure.
+        """
+
+        ret = self.get(name)
+        if ret is None:
+            raise KeyError("failed to find export {}".format(name))
+        return ret
+
+    def get(self, name):
         """
         Looks up an export with `name` on the calling module.
 
--- a/wasmtime/_instance.py
+++ b/wasmtime/_instance.py
@@ -41,6 +41,7 @@ class Instance(object):
             raise Trap.__from_ptr__(trap)
         self.__ptr__ = instance
         self._module = module
+        self._exports = None
 
     @classmethod
     def __from_ptr__(cls, ptr, module):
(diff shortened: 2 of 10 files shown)


alexcrichton:
cc @whitequark, mind taking a look at this?

wasmtime/_func.py
```diff
@@ -141,7 +141,13 @@ class Caller(object):
     def __init__(self, ptr):
         self.__ptr__ = ptr
 
-    def get_export(self, name):
+    def __getitem__(self, name):
+        ret = self.get(name)
+        if ret is None:
+            raise IndexError("failed to find export {}".format(name))
```
whitequark, line 147: So, Python has a `LookupError` exception with two subclasses, `IndexError` and `KeyError`. `IndexError` is for, well, indexes (i.e. integers pointing into sequences), `KeyError` is for keys (any sorts of objects used to look up mappings).

This one should be `KeyError`.

wasmtime/_func.py
```diff
@@ -141,7 +141,13 @@ class Caller(object):
     def __init__(self, ptr):
         self.__ptr__ = ptr
 
-    def get_export(self, name):
+    def __getitem__(self, name):
```
whitequark, line 144: Also it would be nice to have a docstring explaining how this method is different from `get` and which exact exception it raises. Although it does follow the convention, not all libraries do.

wasmtime/_instance.py
```diff
@@ -49,40 +50,59 @@ def __from_ptr__(cls, ptr, module):
...
+        self.extern_list = extern_list
+        self.extern_map = {}
+        exports = module.exports
+        for i, extern in enumerate(extern_list):
+            self.extern_map[exports[i].name] = extern
+
+    def __getitem__(self, idx):
+        ret = self.get(idx)
+        if ret is None:
+            raise IndexError("failed to find export {}".format(idx))
```
whitequark, line 89: Same comments as above regarding the exception, but this case is more complex because there are two different paths (as this object is both a sequence and a mapping).

wasmtime/_instance.py
```diff
@@ -49,40 +50,59 @@ def __from_ptr__(cls, ptr, module):
...
 
     def __del__(self):
         if hasattr(self, '__ptr__'):
             dll.wasm_instance_delete(self.__ptr__)
 
 
+class InstanceExports(object):
+    def __init__(self, extern_list, module):
+        self.extern_list = extern_list
+        self.extern_map = {}
```
whitequark, line 81: I'd probably make these members `_private_ just in case the implementation changes in the future (exposing them doesn't provide any new capability in the API AFAICT).

alexcrichton:
Thanks for the comments!

PR #14, opened 2020-05-22
Remove Python 2-isms

willkg:
This fixes `python_requires` in `setup.py` to be `>= 3.5` since that's what's tested and supported. I don't know how this fix affects things, but it might cause problems when someone with Python 2 tries to install wasmtime-py and it might cause problems with wheel names.

In Python 3, classes don't need to subclass object--all classes are new-style now. This is cosmetic and doesn't change the behavior of anything.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setuptools.setup(
     },
     packages=['wasmtime'],
     include_package_data=True,
-    python_requires='>=2.7',
+    python_requires='>=3.5',
     test_suite="tests",
     extras_require={
         'testing': [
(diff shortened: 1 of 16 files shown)


willkg:
Interesting. Would it be helpful to have installation fail with more fireworks in Python 2 environment?

alexcrichton:
I personally think it's ok for this library to only support Python 3+ so I'm gonna go ahead and merge this. If it turns out we need better error messages for attempts on Python 2 I'm ok crossing that bridge when we get there!

PR #23, opened 2020-05-26
Update how wheels are produced

alexcrichton:
* Produce all wheels on one builder
* Produce an "any" wheel as a catch-all platform
* Rename where libraries are located - allows the "catch-all wheel" to have multiple precompiled libraries
* Fixup CI in a number of ways


whitequark:
Nice, thank you for working on this!

PR #25, opened 2020-05-27
Add type annotations almost everywhere

tobych:
I've added a few missing pointer types to _ffi.py to help here.

I've also added the awkwardly-named "_typing.py", with type aliases:
 
```
XType = typing.Union[FuncType, TableType, MemoryType, GlobalType]
XInstance = typing.Union[Func, Table, Memory, Global]
```

**The module should be called something else.**

**I'm also not sure what the type aliases should be called. I'm new
to WebAssembly and even reading some of the spec, the respective
unions of these don't seem to have names.**

I've run the unit tests with [typeguard](https://typeguard.readthedocs.io/en/latest/) enabled, but typeguard doesn't
switch on `typing.TYPE_CHECKING`. I can sort that out later on.

### DIFF
--- a/wasmtime/_error.py
+++ b/wasmtime/_error.py
@@ -1,4 +1,8 @@
 from ctypes import byref
+import typing
+
+if typing.TYPE_CHECKING:
+    from ._ffi import P_wasmtime_error_t
 
 
 class WasmtimeError(Exception):
@@ -6,7 +10,7 @@ class WasmtimeError(Exception):
         self.message = message
 
     @classmethod
-    def __from_ptr__(cls, ptr):
+    def __from_ptr__(cls, ptr: "P_wasmtime_error_t") -> "WasmtimeError":
         from ._ffi import P_wasmtime_error_t, wasm_byte_vec_t, dll  # Avoid circular import
         if not isinstance(ptr, P_wasmtime_error_t):
             raise TypeError("wrong pointer type")
--- a/wasmtime/_exportable.py
+++ b/wasmtime/_exportable.py
@@ -1,0 +1,12 @@
+__all__ = (
+    "Exportable",
+)
+
+import typing
+
+from ._func import Func
+from ._globals import Global
+from ._memory import Memory
+from ._table import Table
+
+Exportable = typing.Union[Func, Table, Memory, Global]
--- a/wasmtime/_exportable.py
+++ b/wasmtime/_exportable.py
@@ -1,7 +1,3 @@
-__all__ = (
-    "Exportable",
-)
-
 import typing
 
 from ._func import Func
--- a/wasmtime/_extern.py
+++ b/wasmtime/_extern.py
@@ -1,5 +1,10 @@
 from ._ffi import *
+
 from ctypes import *
+import typing
+
+if typing.TYPE_CHECKING:
+    from ._exportable import Exportable
 
 dll.wasm_extern_as_func.restype = P_wasm_func_t
 dll.wasm_extern_as_table.restype = P_wasm_table_t
@@ -8,7 +13,7 @@ dll.wasm_extern_as_memory.restype = P_wasm_memory_t
 dll.wasm_extern_type.restype = P_wasm_externtype_t
 
 
-def wrap_extern(ptr, owner):
+def wrap_extern(ptr: P_wasm_extern_t, owner) -> "Exportable":
     from wasmtime import Func, Table, Global, Memory
 
     if not isinstance(ptr, P_wasm_extern_t):
@@ -33,7 +38,7 @@ def wrap_extern(ptr, owner):
     return Memory.__from_ptr__(val, owner)
 
 
-def get_extern_ptr(item):
+def get_extern_ptr(item: "Exportable") -> P_wasm_extern_t:
     from wasmtime import Func, Table, Global, Memory
 
     if isinstance(item, Func):
@@ -49,7 +54,7 @@ def get_extern_ptr(item):
 
 
 class Extern:
-    def __init__(self, ptr):
+    def __init__(self, ptr: P_wasm_extern_t):
         self.ptr = ptr
 
     def __del__(self):
--- a/wasmtime/_module.py
+++ b/wasmtime/_module.py
@@ -1,6 +1,7 @@
 from ._ffi import *
 from ctypes import *
 from wasmtime import Store, wat2wasm, ImportType, ExportType, WasmtimeError
+import typing
 
 dll.wasmtime_module_new.restype = P_wasmtime_error_t
 dll.wasmtime_module_validate.restype = P_wasmtime_error_t
@@ -8,7 +9,7 @@ dll.wasmtime_module_validate.restype = P_wasmtime_error_t
 
 class Module:
     @classmethod
-    def from_file(cls, store, path):
+    def from_file(cls, store: Store, path):
         """
         Compiles and creates a new `Module` by reading the file at `path` and
         then delegating to the `Module` constructor.
@@ -18,7 +19,7 @@ class Module:
             contents = f.read()
         return cls(store, contents)
 
-    def __init__(self, store, wasm):
+    def __init__(self, store: Store, wasm: typing.Union[str, bytes]):
         if not isinstance(store, Store):
             raise TypeError("expected a Store")
 
@@ -46,7 +47,7 @@ class Module:
         self.store = store
 
     @classmethod
-    def validate(cls, store, wasm):
+    def validate(cls, store: Store, wasm: typing.Union[bytes, bytearray]) -> None:
         """
         Validates whether the list of bytes `wasm` provided is a valid
         WebAssembly binary given the configuration in `store`
@@ -68,7 +69,7 @@ class Module:
             raise WasmtimeError.__from_ptr__(error)
 
     @property
-    def imports(self):
+    def imports(self) -> typing.List[ImportType]:
         """
         Returns the types of imports that this module has
         """
@@ -81,7 +82,7 @@ class Module:
         return ret
 
     @property
-    def exports(self):
+    def exports(self) -> typing.List[ExportType]:
         """
         Returns the types of the exports that this module has
         """
--- a/wasmtime/_types.py
+++ b/wasmtime/_types.py
@@ -1,5 +1,9 @@
 from ._ffi import *
 from ctypes import *
+import typing
+
+if typing.TYPE_CHECKING:
+    from ._value import P_wasm_valtype_vec_t
 
 dll.wasm_valtype_new.restype = P_wasm_valtype_t
 dll.wasm_functype_new.restype = P_wasm_functype_t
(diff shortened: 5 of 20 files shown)


tobych:
Regarding `XInstance`, above. Found it. Section 2.5.10 of the WebAssembly Specification, Release 1.1:

> Each export is labeled by a unique _name_. Exportable definitions are _functions_, _tables_, _memories_, and _globals_, which are referenced through a respective descriptor.

So I'll rename `XInstance` so we have:

```
ExportableDefinition = typing.Union[Func, Table, Memory, Global]
```

alexcrichton:
Thanks for this! I'm pretty new to Python type annotations, but is there a way that we could add some more CI here? Some things I think would be neat would be:

* Can we require that all functions in `wasmtime` have type annotations?
* Can we enable type checking in CI while running the test suite?

tobych:
@alexcrichton Yes, Python package [typeguard](https://pypi.org/project/typeguard/) can check arguments and results of function calls at runtime. It has a pytest [plugin](https://typeguard.readthedocs.io/en/latest/#using-the-pytest-plugin), which I've been using to work on this ticket, like this:

```sh
$ pytest --typeguard-packages=wasmtime
```

Also, that's another +1 for pytest, I guess.

I've not been able to use it much though, because there are so many circular imports in the package currently. That'll be helped with the `__all__` work in #17, but might also need explicit imports rather than `import *`. It will also need the `TYPE_CHECKING` test to be replace with logic that checks to see if typeguard is running. I've done something like that in another (commercial) project and it worked well.

As far as I know, typeguard can't check to see if every function has type annotations for all arguments. I doubt that would be worth doing, not least because it might be a step too far to developers contributing to the project, but I've not thought it through.

tobych:
I'll rename `XType` to `ExternalType`, to match section 2.3.8 of the WebAssembly Specification:

image

tobych:
Okay @alexcrichton, this is ready to merge. Could be maybe.

wasmtime/_error.py
```diff
@@ -1,12 +1,16 @@
 from ctypes import byref
+import typing
+
+if typing.TYPE_CHECKING:
+    from ._ffi import P_wasmtime_error_t
```
alexcrichton, line 5: How come this is conditional?

wasmtime/_error.py
```diff
@@ -1,12 +1,16 @@
...
 
 
 class WasmtimeError(Exception):
     def __init__(self, message):
         self.message = message
 
     @classmethod
-    def __from_ptr__(cls, ptr):
+    def __from_ptr__(cls, ptr: "P_wasmtime_error_t") -> "WasmtimeError":
```
alexcrichton, line 13: I'm not super familiar with this, but how come these are strings?

wasmtime/_exportable.py
```diff
@@ -0,0 +1,12 @@
...
+)
+
+import typing
+
+from ._func import Func
+from ._globals import Global
+from ._memory import Memory
+from ._table import Table
+
+Exportable = typing.Union[Func, Table, Memory, Global]
```
alexcrichton, line 12: To match the naming in wasmtime itself, could this be named `Extern`?

wasmtime/_exportable.py
```diff
@@ -0,0 +1,12 @@
+__all__ = (
+    "Exportable",
+)
```
alexcrichton, line 3: Could this avoid the usage of `__all__` while the other PR is sorted out?

wasmtime/_extern.py
```diff
@@ -1,5 +1,10 @@
 from ._ffi import *
+
 from ctypes import *
+import typing
+
+if typing.TYPE_CHECKING:
+    from ._exportable import Exportable
```
alexcrichton, line 7: I think it's fine to do this unconditionally? Or am I missing something for why we might want this to not happen all the time?

wasmtime/_exportable.py
alexcrichton: Ah I see this conflicts with the literal `Extern` type. Could we take a leaf out of Rust's book and do something like `IntoExtern`?

wasmtime/_module.py
```diff
@@ -1,14 +1,15 @@
...
+import typing
 
 dll.wasmtime_module_new.restype = P_wasmtime_error_t
 dll.wasmtime_module_validate.restype = P_wasmtime_error_t
 
 
 class Module:
     @classmethod
-    def from_file(cls, store, path):
+    def from_file(cls, store: Store, path):
```
alexcrichton, line 12: I think `path` here is missing a type annotation?

alexcrichton: (is there a way to catch this?)

wasmtime/_types.py
```diff
@@ -425,3 +429,6 @@ def type(self):
     def __del__(self):
         if self.__owner__ is None:
             dll.wasm_exporttype_delete(self.__ptr__)
+
+
+ExternalType = typing.Union[FuncType, TableType, MemoryType, GlobalType]
```
alexcrichton, line 434: Could this use similar naming to the other union? Something like `IntoExternType`?

wasmtime/_value.py
```diff
@@ -1,10 +1,11 @@
 from ._ffi import *
 from wasmtime import ValType
+import typing
 
 
 class Val:
     @classmethod
-    def i32(cls, val):
+    def i32(cls, val: int) -> "Val":
```
alexcrichton, line 8: This seems like a pretty common pattern we have, where within a `class` definition there's methods that take/return that class. Is there really no other way to express this than with a string?

wasmtime/_wasi.py
```diff
@@ -109,7 +110,7 @@ def __init__(self, store, name, config):
         self.__ptr__ = ptr
         self.store = store
 
-    def bind(self, import_):
+    def bind(self, import_: ImportType) -> typing.Optional["XInstance"]:
```
alexcrichton, line 113: Looks like `XInstance` is still mentioned here?

(how come this wasn't caught through testing?)

wasmtime/_error.py
tobych: I'm avoiding importing anything into the namespace just for type-checking.

wasmtime/_error.py
tobych: Strings because they're only imported when type checking (as per TYPE_CHECKING). The imports aren't needed at run-time. The strings mean Python will ignore them at run-time. It looks at the string when it's doing static type checking though.

wasmtime/_exportable.py
tobych: I'll look into that. So many things called the same thing... Trying to correlate the WebAssembly spec, undocumented C API, wasmtime API and wasmtime-py names has been... a journey. I see a spreadsheet in my future.

wasmtime/_exportable.py
tobych: Sure.

wasmtime/_extern.py
tobych: Sure. To explain, I treat every import into any namespace as potentially breaking things, or potentially bringing in tonnes of code, without realizing it, that doesn't need to be there at run-time. I'm used to working on large, monolithic machine learning codebases where one unnecessary import could bring Pandas into a small Docker container when it's not needed.

wasmtime/_module.py
tobych: I've not tried to annotate every type, actually. Trying to do that with Python, you'll end up hitting a point of diminishing returns. `path` can be things other than a `str` here but I wasn't sure what to do. `str` would be okay.

wasmtime/_value.py
tobych: The quotes are there because `Val` isn't yet known to Python as it's compiling the definition of the class. It's a [forward-reference](https://www.python.org/dev/peps/pep-0484/#id36).

wasmtime/_wasi.py
tobych: Remember nothing is being checked at runtime by Python. And getting `typeguard`, the package we could use to do some run-time checking of some things, working with all the `import *` and so on was too much. I figured we could get `typeguard` on board once the `import *` have all gone away.

tobych:
On a wider note, with regard to questions about CI, I think doing run-time type checking in CI is likely to be of use only if we can add pragmas to the code to suppress complaints. And probably as a report rather than a gate.

Even then, I'm realizing there's way more run-time type-checking in this package than I've seen in other Python code. Maybe it's insufficient experience on my part (but I've been using Python for 15 years), but I'm used to leaving type-checking to static analysis (in my IDE), and type-checking arguments using _typeguard_ on specific functions (with its decorators). All the type-checking makes the code hard to read. And there are even unit tests checking that the type-checking code is working, which take up even more lines of code. Code that currently fails type-checking tests, of course. I get that test coverage is important, but something like _typeguard_ can do that type-checking for free (code lines, not time; run-time type-checking can take a lot of time), just by adding a decorator to a function. That way you don't need to test type-checking in individual functions: you can just trust _typeguard_ to do it for you.

On a side note, I'm finding this totally fascinating, realizing I'm a Python developer even more than I'd realized, and a C programmer or Rust programmer (not least one on the core team, as Alex is) will be coming to Python with so many other habits and values. Personally I'm learning Go, not least because I've got too used to Python. I'm now learning Rust, too, not least because so much code around this new wonderful world of WebAssembly is written in Rust and I need to be able to read it. It's also going to throw me back into the world of compile-time type-checking that won't actually let you do it. Something every Python programmer should dip into on occasion.

wasmtime/_types.py
tobych: - [ ] I'll look into this.

I'll look at `wasmtime.h` and read up on this Rusty `Into` thing.

Personally (strong opinion, weakly held), I'm very happy having things named the same across different modules. I'm used to naming things super close to what's used in the domain, and if I end up with the same names across modules, I'll just use `from fruit import Banana as Banana_` or `import fruit` then use `fruit.Banana` in the code. I'm already struggling to cope with the names of things in this package not matching those in `wasm.h` or the wasmtime API or the WebAssembly spec as it is: for example, I think `Instance` should be called `ModuleInstance`, since there are _instances_ of several other things in the runtime environment. So starting to add `Into` at the start might be a move too far. :-)

wasmtime/_extern.py
alexcrichton: That makes sense, but this is all code controlled in this package, so I think we can ensure that we keep a handle on our own imports?

wasmtime/_types.py
alexcrichton: Er well in any case I think the naming here and the actual instances themselves should be the same. For example `External` and `ExternalType`

wasmtime/_value.py
alexcrichton: That makes sense I'm just surprised that such a common pattern doesn't have better support...

alexcrichton:
I'm not a huge fan of all the type checking here either, but it's required for the semantics I was shooting for. If the type checking is removed then it's very easy to segfault this extension because the wrong type was passed around by accident. I basically want to guarantee that any buggy usage results in exceptions, not segfaults, so type checking was required.

I'm pretty hesitant to merge this without any CI checking the annotations. Once there's CI I think it's fine, but this should be checked *somehow* in an automated way.

alexcrichton:
I poked a bit more at this today to learn a bit more about it. It looks like `mypy` is a good candidate for running tests. I edited `pytest.ini` to add `--mypy` and installed `pytest-mypy` as well. Unfortunately it prints out a pretty huge number of type errors.

I don't mind fixing all the issues, but I'm pretty stuck when it comes to ctypes. All lines [like this](https://github.com/bytecodealliance/wasmtime-py/blob/e389990ce32e9ea4434ae5f098dd2f81d78b1ce2/wasmtime/_globals.py#L5-L8) generate errors which are probably fine to ignore with `# type: ignore`, but all *invocations* of functions from the loaded library (e.g. `dll.wasm_*`) don't typecheck since mypy can't figure out the types there.

I'm stuck on how to structure all this well. Ideally mypy would be told what the type signature is for all functions. I'm wondering if this might be best to do a code generation sort of thing where we generate `_ffi.py` from all the header files...

alexcrichton:
Merged with https://github.com/bytecodealliance/wasmtime-py/pull/29, thanks again for the work here @tobych!

PR #35, opened 2020-06-23
Bind `wasmtime_trap_exit_status`

### LINKED ISSUE #30: Handling the exit trap
Since wasmtime-py 0.18.0, it seems that when the `main()` function returns, the call to `_start()` also simply returns, which is good. However, it looks like if `exit()` is called, then the result is a `wasmtime._trap.Trap: Exited with i32 exit status 1`. It is not clear how this trap can be handled besides matching the exception message with a regex.

alexcrichton:
* Add an `ExitTrap` subclass of `Trap`
* Return an appropriate trap from `Trap.__from_ptr__`
* Attach the exit code via a `code` field

Closes #30

### DIFF
--- a/tests/test_trap.py
+++ b/tests/test_trap.py
@@ -78,3 +78,32 @@ wasm backtrace:
         self.assertEqual(frames[0].func_index, 0)
         self.assertEqual(frames[0].func_name, None)
         self.assertEqual(frames[0].module_name, None)
+
+    def test_wasi_exit(self):
+        store = Store()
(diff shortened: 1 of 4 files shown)


whitequark:
Thank you! Can I get a release? The lack of exit trap handling is currently making all of my wasmtime-py based applications crash on exit, so not very good.

alexcrichton:
Sure thing! I updated to 0.18.1

whitequark:
Thanks, works well

PR #38, opened 2020-07-13
Split contribution instructions out to CONTRIBUTING.md

fitzgen:
Also update the set up and testing instructions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,16 +4,15 @@
 Alliance's [Code of Conduct] and [Organizational Code of Conduct].
 
 So far this extension has been written by folks who are primarily Rust
-programmers, so it's highly likely that there's some faux pas in terms of Python
-idioms. Feel free to create a PR to help make things more idiomatic if you see
-something!
+programmers, so feel free to create a PR to help make things more idiomatic if
+you see something!
 
 ## Set Up
 
 You'll need to acquire a [Wasmtime] installation. The `wasmtime-py` package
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,65 @@
+# Contributing to `wasmtime-py`
+
+`wasmtime-py` is a [Bytecode Alliance] project. It follows the Bytecode
+Alliance's [Code of Conduct] and [Organizational Code of Conduct].
+
+So far this extension has been written by folks who are primarily Rust
+programmers, so it's highly likely that there's some faux pas in terms of Python
+idioms. Feel free to create a PR to help make things more idiomatic if you see
+something!
```
whitequark, line 9: I'm pretty sure that by this point, this extension does qualify as "idiomatic Python"!

fitzgen: Heh yeah I'm happy to remove this clause if its unnecessary now (I'm also not really in a position to judge whether this package is idiomatic or not :sweat_smile: )

whitequark: I'm sure there are a few things that could be further improved, but Python packages aren't all that consistent in general. Wasmtime-py seems at the very least better than average to me, which means the warning no longer applies!

alexcrichton:
I've pushep up a few wording tweaks which makes the ask for help on idioms a bit less zealous (but still there because I'm paranoid), and I've also freshened up the instructions for handling the native lib.

PR #39, opened 2020-07-13
Rename `Val.__raw__` to `Val._raw`

fitzgen:
Stylistically we shouldn't use double under for non-builtin fields.

### DIFF
--- a/wasmtime/_config.py
+++ b/wasmtime/_config.py
@@ -21,7 +21,7 @@ class Config:
     """
 
     def __init__(self) -> None:
-        self.__ptr__ = ffi.wasm_config_new()
+        self._ptr = ffi.wasm_config_new()
 
     @setter_property
     def debug_info(self, enable: bool) -> None:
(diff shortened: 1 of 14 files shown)


whitequark:
(It's not just style, right? The way dunder fields are treated in Python is roughly equivalent to how defining a function called `is\w` in C is undefined behavior.)

fitzgen:
> (It's not just style, right? The way dunder fields are treated in Python is roughly equivalent to how defining a function called `is\w` in C is undefined behavior.)

Perhaps! Its been about a decade since I seriously wrote any python, other than for little one-off scripts.

whitequark:
In theory, CPython can add new ones at any time and so break any user code that used these for its own purposes. In practice there seems to be an unspoken agreement that libraries are allowed to grab names from that namespace in exceptional cases, but I'd still just treat them all as potential future UB.

fitzgen:
I see there are a few other similar things, so I will fix those in another commit here.

alexcrichton:
Ah yeah I think this is me misunderstanding things, and at least the doc generators used here recognizes "leading underscore is private" which is the intention for all these fields anyway, so sounds good. Thanks @fitzgen!

PR #41, opened 2020-07-16
Expose reference types in the Python API

fitzgen:
~~Depends on https://github.com/bytecodealliance/wasmtime/pull/2029~~ This merged.

~~Does not pull in changes required for https://github.com/bytecodealliance/wasmtime/pull/2026 yet.~~ Does now.

### DIFF
--- a/wasmtime/_ffi.py
+++ b/wasmtime/_ffi.py
@@ -2,6 +2,7 @@ from ctypes import *
 import os
 import sys
 import platform
+import typing
 
 from wasmtime import WasmtimeError
 
@@ -39,18 +40,24 @@ WASM_CONST = c_uint8(0)
 WASM_VAR = c_uint8(1)
 
 
+class wasm_ref_t(Structure):
+    pass
+
+
 class wasm_val_union(Union):
     _fields_ = [
         ("i32", c_int32),
         ("i64", c_int64),
         ("f32", c_float),
         ("f64", c_double),
+        ("ref", POINTER(wasm_ref_t)),
     ]
 
     i32: int
     i64: int
     f32: float
     f64: float
+    ref: "typing.Union[pointer[wasm_ref_t], None]"
 
 
 class wasm_val_t(Structure):
--- a/wasmtime/_func.py
+++ b/wasmtime/_func.py
@@ -44,7 +44,7 @@ class Func:
         self._owner = None
 
     @classmethod
-    def __from_ptr__(cls, ptr: "pointer[ffi.wasm_func_t]", owner: Optional[Any]) -> "Func":
+    def _from_ptr(cls, ptr: "pointer[ffi.wasm_func_t]", owner: Optional[Any]) -> "Func":
         ty: "Func" = cls.__new__(cls)
         if not isinstance(ptr, POINTER(ffi.wasm_func_t)):
             raise TypeError("wrong pointer type")
@@ -58,7 +58,7 @@ class Func:
         Gets the type of this func as a `FuncType`
         """
         ptr = ffi.wasm_func_type(self._ptr)
-        return FuncType.__from_ptr__(ptr, None)
+        return FuncType._from_ptr(ptr, None)
 
     @property
     def param_arity(self) -> int:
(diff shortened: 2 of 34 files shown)


wasmtime/_table.py
```diff
@@ -53,65 +44,75 @@ def type(self) -> TableType:
...
-        be an integer index.
+        The `val` specified must be convertible into this table's element
+        type. I.e. for a `funcref` table, `val` must either be a `Func` or
+        `None`, and for an `externref` table, `val` may be any arbitrary
+        external data.
 
         Raises a `WasmtimeError` if `idx` is out of bounds.
         """
+        if idx >= self.size:
+            raise WasmtimeError("Index out of bounds when setting table element")
```
alexcrichton, line 110: This should be part of the `wasm_table_set` API I think?

alexcrichton: Or if this is for a better error message perhaps this could happen in the `not ok` branch?

wasmtime/_table.py
```diff
@@ -1,44 +1,35 @@
...
 
-        init_ptr = get_func_ptr(init)
-        ptr = POINTER(ffi.wasm_table_t)()
-        error = ffi.wasmtime_funcref_table_new(store._ptr, ty._ptr, init_ptr, byref(ptr))
-        if error:
-            raise WasmtimeError.__from_ptr__(error)
+        if init is None:
+            init_val = Val.ref_null(ty.element)
+        else:
+            init_val = Val._convert(ty.element, init)
```
alexcrichton, line 23: Technically this branch isn't necessary, right? This is already handled by `_convert`?

wasmtime/_func.py
```diff
@@ -201,20 +209,22 @@ def invoke(idx, params_ptr, results_ptr, params):  # type: ignore
...
+            val = Val._convert(result_tys[0], results)
+            val._owner = "passing ownership into Wasm"
             results_ptr[0] = val._raw
         else:
             if len(results) != len(result_tys):
                 raise WasmtimeError("callback produced wrong number of results")
             for i, result in enumerate(results):
-                val = Val.__convert__(result_tys[i], result)
+                val = Val._convert(result_tys[i], result)
+                val._owner = "passing ownership into Wasm"
```
alexcrichton, line 227: Instead of setting the `_owner` here so it doesn't do anything in the destructor, could this instead perhaps do something like `take()` where it cancels the destructor internally? I think ideally only `_val.py` should be reaching into the internal fields of `Val`

wasmtime/_func.py
```diff
@@ -201,20 +209,22 @@ def invoke(idx, params_ptr, results_ptr, params):  # type: ignore
 
     try:
         for i in range(0, len(param_tys)):
-            params.append(extract_val(Val(params_ptr[i])))
+            params.append(Val(params_ptr[i]).value)
```
alexcrichton, line 212: I think this may not be quite right because we're not given ownership of parameters?

wasmtime/_func.py
```diff
@@ -91,11 +91,15 @@ def __call__(self, *params: IntoVal) -> Union[IntoVal, Sequence[IntoVal], None]:
 
         ty = self.type
         param_tys = ty.params
+        if len(params) > len(param_tys):
+            raise WasmtimeError("too many parameters provided")
+        if len(params) < len(param_tys):
+            raise WasmtimeError("too few parameters provided")
+
+        param_vals = [Val._convert(ty, params[i], "keep alive across call")
```
alexcrichton, line 99: FWIW there's *tons* of instances where we don't keep something alive across a call, but I'm led to believe that's ok in Python because reference counting doesn't happen until the end of a scope. In that sense I don't think that we need to manually set an owner here to keep things alive across the call.

wasmtime/_ffi.py
```diff
@@ -39,18 +39,24 @@
...
         ("f32", c_float),
         ("f64", c_double),
+        ("ref", POINTER(wasm_ref_t)),
     ]
 
     i32: int
     i64: int
     f32: float
     f64: float
+    ref: POINTER(wasm_ref_t)  # type: ignore
```
alexcrichton, line 59: I think this may need to be `"pointer[wasm_ref_t]"` (don't ask me why)

wasmtime/_table.py
fitzgen: Yeah, the `wasm_table_get` signature doesn't differentiate between kinds of failures or provide any sort of message, so this is doing an explicit check to provide the better error message.

fitzgen:
Alright, I think everything you mentioned in review should be taken care of. `Val` is essentially an `Option<wasm_val_t>` now, where passing ownership into wasm sets the raw value to `None`.

PR #44, opened 2020-08-25
Fix wasmtime.loader to import module using wasi_snapshot_preview1

tkat0:
Hi @alexcrichton.

I found that `wasmtime.loader` can't import the module importing `wasi_snapshot_preview1`

```python
# load_test.py
import wasmtime.loader
import hello_wasi # module using wasi_snapshot_preview1
```

```bash
$ python load_test.py
...
load_test.py:2: in <module>
    import hello_wasi
wasmtime/loader.py:71: in exec_module
    imported_module = importlib.import_module(module_name)
../.pyenv/versions/3.8.2/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ModuleNotFoundError: No module named 'wasi_snapshot_preview1'
```

Currently, the loader can import any python modules, but loader must skip predefined modules in this library.

How about this fix?

### DIFF
--- a/examples/loader.py
+++ b/examples/loader.py
@@ -5,6 +5,7 @@ import wasmtime.loader  # noqa: F401
 
 import loader_load_python  # type: ignore
 import loader_load_wasm  # type: ignore
+import loader_load_wasi  # type: ignore # noqa: F401
 
 # This imports our `loader_add.wat` file next to this module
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

PR #45, opened 2020-09-19
Fixed Python hello example on README.md

viniarck:
When I first tried out the example in the README.md, I hit this `TypeError` exception: 

```
  File "ex.py", line 4, in <module>
    module = Module(store, """
  File "repos/wasmtime-py/wasmtime/_module.py", line 21, in __init__
    raise TypeError("expected an Engine")
TypeError: expected an Engine
```

The `Module` class used to take a `Store` but it has changed to take a `Engine` instead, so I've adapted accordingly. If I run this script (copying and pasting from the README.md example) I get the expected outcome: 

```
❯ python script.py
Hello from Python!
```

Awesome project by the way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,11 @@ macOS, and Linux
 In this example, we compile and instantiate a WebAssembly module and use it from Python:
 
 ```python
-from wasmtime import Store, Module, Instance, Func, FuncType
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

PR #66, opened 2021-08-18
Fix `wasmtime.loader` with multiple functions

### LINKED ISSUE #65: wrong function called from module when using wasmtime loader
using `wasm_module.wat`:
```
(module
  (type (;0;) (func (result i32)))
  (func $func1 (type 0) (result i32)
    i32.const 1)
  (func $func2 (type 0) (result i32)
    i32.const 2)
  (table (;0;) 1 1 funcref)
  (memory (;0;) 16)
  (global (;0;) (mut i32) (i32.const 1048576))
  (global (;1;) i32 (i32.const 1048576))
  (global (;2;) i32 (i32.const 1048576))
  (export "memory" (memory 0))
  (export "func1" (func $func1))
  (export "func2" (func $func2))
  (export "__data_end" (global 1))
  (export "__heap_base" (global 2)))
```

when loaded as a python module, calling `func1` returns 2 instead of 1.
this is because `func2` is called instead, can be verified by modifying `func2`

tested with this script:
```
from wasmtime import Store, Module, Instance, Func, FuncType, loader
import wasm_module

store = Store()
module = Module(store.engine, open("wasm_module.wat").read())

instance = Instance(store, module, [])
func1 = instance.exports(store)["func1"]
assert func1(store) == wasm_module.func1(), f"from instance {func1(store)}, from loader {wasm_module.func1()}"
```

the error message is 
```
  File "/home/user/Documents/test/test.py", line 9, in <module>
    assert func1(store) == wasm_module.func1(), f"from instance {func1(store)}, from loader {wasm_module.func1()}"
AssertionError: from instance 1, from loader 2
```

tested with `wasmtime-0.29.0.dev168`

alexcrichton:
This is showing off how I'm not exactly a Python expert, but the
previous iteration didn't work because lambda closure captures didn't
work as I expected!

Closes #65

### DIFF
--- a/wasmtime/loader.py
+++ b/wasmtime/loader.py
@@ -86,7 +86,7 @@ class _WasmtimeLoader(Loader):
             # to our loader's store
             if isinstance(item, Func):
                 func = item
-                item = lambda *args: func(store, *args)  # noqa
+                item = lambda *args,func=func: func(store, *args)  # noqa
             module.__dict__[export.name] = item
 
 


alexcrichton:
The idiom here is drawing inspiration from https://stackoverflow.com/questions/2295290/what-do-lambda-function-closures-capture/2295372#2295372 for what to do.

PR #85, opened 2022-03-24
Remove support for module linking

alexcrichton:
This gets tests to pass now that bytecodealliance/wasmtime#3958 has
landed.


alexcrichton:
Note that this is dependent on https://github.com/bytecodealliance/wasmtime/pull/3959 to work

PR #99, opened 2022-10-24
Bump to 2.0.0

alexcrichton:


whitequark:
Do you think there's going to be actual breaking changes post-1.0.0? Proper semver-style dependencies are a significant maintenance burden for my packages that depend on wasmtime-py because of the dependency resolution conflicts they create (as pip cannot install wasmtime-1.0.0 and wasmtime-2.0.0 concurrently).

alexcrichton:
The [original RFC for versioning](https://github.com/bytecodealliance/rfcs/blob/main/accepted/wasmtime-one-dot-oh.md) goes into some more details, but the general gist is that the Rust crate plans to maintain the bump-the-major-version cadence of versioning and to make it easy to determine what version of wasmtime you're using we decided to keep all language bindings at the same version. This prevents us from having to maintain a complicated cadence of what-language-binding-is-at-what-version in addition to evaluating each language binding each release to see if breaking changes were made.

That being said this was largely developed from a Rust point of view where having two major versions is not that big of an issue. I was unaware of pip's requirement that it can't have 1.0 and 2.0 simultaenously. Is this versioning scheme predicted to create issues for Python?

whitequark:
> Is this versioning scheme predicted to create issues for Python?

It is already creating issues; for example, [YoWASP](https://yowasp.org) packages and [amaranth-yosys](https://github.com/amaranth-lang/amaranth-yosys/) packages, which are released at a different cadence, routinely create intractable dependency conflicts due to varying semver-style requirements on wasmtime. To make matters worse, while pip's dependency resolution has improved recently, even a situation like the following poses problems:
  - yowasp-yosys depending on wasmtime>=1.0.0,<3.0
  - amaranth-yosys depending on wasmtime>=1.0.0,<2.0
 In this case, if you attempt to install the two packages at once (whether explicitly or as a transitive dependency), pip will choose wasmtime==2.0 and then fail since amaranth-yosys' dependency is unsatisfied, in spite of there being a valid solution (choosing wasmtime==1.0).

alexcrichton:
Hm ok, am I correct in that the only way to improve the situation is to not release a second major version? Or otherwise is there anything else that can be done on our end?

whitequark:
An ideal situation is one where a breaking semver change happens only when the API is actually broken from a consumer's point of view. So long as it happens every month, I have one of the two options:
 * Remove the upper limit on wasmtime dependency entirely. This means that when you release a breaking change, all of my packages will inevitably break.
 * Keep releasing new versions of my packages every month, mostly just updating the upper bound. This means added workload, and inevitably, friction for downstream users when for some reason I'm not able to do so.

whitequark:
I am currently following the latter option, but because of said friction and maintenance burden, I'm leaning towards the former option. If I pick it, I will have to tell downstream users of my software "if you install one of my packages and it doesn't work, it happened because of wasmtime-py's versioning policy, and I cannot necessarily tell you which version you need to install for your entire depdendency set to function correctly, or whether there is such a version".

alexcrichton:
Hm ok, I'm sorry for the friction here, this is not something I think any of us knew would be a consequence of the versioning scheme. I don't really have a great answer for you at this time, so I would agree that if your goal is to reduce your own maintenance then having an open-ended version req is probably best for now.

We in general I think at this time don't have the resources to meticulously maintain each language's bindings and make per-language evaluations for each release as to whether a breaking change was present. Additionally we didn't want to have to deal with triaging reports about figuring out what version of wasmtime is actually in use. All that to say that the versioning scheme is currently relatively heavily skewed in the favor of wasmtime developers, which probably isn't the final answer long-term but assists us in the short term of getting more things out there. Not to make excuses, but trying to give some more background on what's at least my point of view.

whitequark:
Got it, I will switch to an open-ended versioning scheme and hope there aren't too many breaking changes. I am also going to add weekly (?) cron-triggered CI builds that verify functionality against latest wasmtime.

alexcrichton:
Ok that sounds good. So far the C API has been fairly stable and no major changes are necessarily planned, but as with anything it's tough to say how long it will be stable for in the long-term. I'll try to make sure to leave any breaking changes open as a PR for awhile to get feedback in the future.

PR #101, opened 2022-11-13
fix: typing for Callable in linker.define_func

Niraj-Kamdar:
Type checker complains the type is partially unknown when the type isn't fully defined. This will fix it.


alexcrichton:
Thanks! If you rebase I think CI should be fixed, and can you also add a test exercising this?

alexcrichton:
Sorry didn't notice this was rebased!

PR #104, opened 2022-11-17
Merge component bindings generation into this repository

alexcrichton:
This is a very large PR which is tasked with, at a high level, merging the `wit-bindgen-gen-host-wasmtime-py` generator from the `wit-bindgen` project into this repository. My purpose for implementing this is to chart a path for moving the Wasmtime and JS host generators out of `wit-bindgen` as well, so while this is Python-specific I'm hoping that all the lessons learned here can be used as inspiration for those migrations as well.

This is, however, not "just a simple merge" of moving code around. The `wit-bindgen` project has its own way of integrating into a CLI and additionally has its own setup for tests. Neither of those fit very well into this repository so I opted to take a different path. What I've ended up settling on feels, I think at least, more natural for Python and more easily integrates with this repository. At a high-level the major points of this movement are:

* The `wit-bindgen-gen-host-wasmtime-py` crate is moved to this repository under a `rust` directory. This is then exported as a `wasmtime.bindgen` module from the `wasmtime` package this repository represents. This is internally implemented by compiling the Rust code itself to WebAssembly and then using the `wasmtime` package itself to run the wasm to power the Python APIs. More-or-less this is a prototype of "what if wasm modules were used as Python extensions?". I've opted to do this since I would like to keep the generation of Python in Rust as all `*.wit` tooling is in Rust. Additionally I don't want to go through dll `ctypes` to integrate everything into Python in the same manner the Wasmtime C API is integrated. Thus, this demos a self-hosted Python bindings generator.
* Script-related things are moved under `ci` such as `ci/download-wasmtime.py` and `ci/cbindgen.py` (renamed from `ci/bindgen.py`)
* A new `ci/build-rust.py` will build the Rust code into a component and generate the `wasmtime.bindgen` bindings to run this component.
* Tests have been added under `tests/codegen/test_*` which will, for specific components, generate bindings and then run them locally.
* Bindings generation is available via a simple program to users through `python -m wasmtime.bindgen the-component.wasm --out-dir foo`. I intentionally haven't wrapped this up in a nice script since I don't have a good name for it. This, however, is intended to be the entry point for documenting how end users can actually use this. As a bonus it doesn't require building a bunch of Rust code as a Python developer.


alexcrichton:
I'm opening this as a draft as mostly a declaration of "this is happening". Otherwise though remaining items for me are:

* Writing more tests
* Write more documentation

The main plumbing of everything should be there and is almost ready and good to go.

PR #124, opened 2023-02-03
Update for the latest C API changes

alexcrichton:


wasmtime/_linker.py
```diff
@@ -32,7 +32,7 @@ def allow_shadowing(self, allow: bool) -> None:
             raise TypeError("expected a boolean")
         ffi.wasmtime_linker_allow_shadowing(self._ptr, allow)
 
-    def define(self, module: str, name: str, item: AsExtern) -> None:
+    def define(self, store: Storelike, module: str, name: str, item: AsExtern) -> None:
```
whitequark, line 35: @alexcrichton This is a breaking change, right? Re: our discussion in #99.

alexcrichton: Ah, yes indeed! Is that ok?

alexcrichton: Ah ok, this is currently in-line with our development policy for Wasmtime. This is why a new major version is released every month instead of minor versions and such, although I understand that the Python packaging situation complicates things.

If you're looking for a more ironclad gaurantee of stability then something like `wasm2c` might work better for you, since at least at this time the `wasmtime-py` embedding can't offer 100% stability.

whitequark: > If you're looking for a more ironclad gaurantee of stability

I don't specifically care about _stability_ so much as I care about my downstream users being able to actually use packages I ship. I feel like as things stand, publishing wasmtime-py wheels on PyPI isn't even that useful beyond a proof of concept.

Potentially pipx may be able to improve things here, I'll look into using that with a more restrictive version constraint.

whitequark: > It sounds, though, like short of publishing `wasmtime4`, `wasmtime5`, ... as package names there may unfortuantely not be much we can do.

That would work in a technical sense but is almost certainly not worth it due to the added friction.

> I've you've got suggestions of how to make wasmtime-on-pypi more useful I'm definitely a novice when it comes to Python so I'd be happy to change things.

How do the other bindings deal with this issue? Is it just through installing multiple versions?

whitequark: Honestly one of the things that might work is re-publishing `wasmtime` wheels under e.g. `yowasp-wasmtime`, with the identical content but different name. That's pretty cursed but it might be the least bad solution here, considering that `pipx` won't work in the cases where `yowasp-yosys`, etc are used as dependencies of something else.

alexcrichton: > How do the other bindings deal with this issue?

So far I think the main language of consumption for Wasmtime is Rust, where Cargo sorts everything out here. I'm not sure how other ecosystems are managing this or whether it's a problem there too.

> Honestly one of the things that might work is re-publishing wasmtime wheels under e.g. yowasp-wasmtime

If you'd like any support code here in this to make the "fork" easier to do on your end I'm happy to merge that in here!

whitequark: I'll take a look at it and get back to you. Thanks!

whitequark: So after a lot of deliberation I think the solution I'll go with is a GHA cronjob that tests the packages against latest wasmtime and if tests pass automatically relaxes the bound on wasmtime. It's a bit cursed but since that's effectively what I've been doing manually I think automating it should work about as well and probably good enough for my use case.

whitequark: Here's my solution: https://github.com/whitequark/track-pypi-dependency-version

I think it might be worth recommending in README or something.

alexcrichton: Sounds reasonable to me! Would you like to send a PR?

whitequark: https://github.com/bytecodealliance/wasmtime-py/pull/125

PR #134, opened 2023-03-05
FIXES #81: 400x faster slice assignment

### LINKED ISSUE #81: Setting a slice of memory without having to enumerate
It's nice that the WASM memory can be inspected with a slice operator:

```python
from wasmtime import *
engine = Engine()
store = Store(engine)
module = Module.from_file(engine, "./hello.wasm")
instance = Instance(store, module, [])
exports = instance.exports(store)
memory = exports["memory"]
```

Then

```python
>>> memory.data_ptr(store)[1024:1029]
[114, 97, 116, 115, 0]
```

but it's not possible to set it that way which seems like a bug, or at least a missing feature:

```python
>>> data[1024:1028] = [114, 97, 116, 116]
  File "<stdin>", line 1, in <module>
TypeError: sequence index must be integer, not 'slice'
```

You can work around it by iterating:

```python
for i,v in enumerate("fats".encode('utf-8')): data[i] = v
```

muayyad-alsadi:
for more information check #81

### DIFF
--- a/wasmtime/_memory.py
+++ b/wasmtime/_memory.py
@@ -62,6 +62,51 @@ class Memory:
         """
         return ffi.wasmtime_memory_data(store._context, byref(self._memory))
 
+    def set_store(self, store: Storelike):
+        """
+        must be called to set the store for highlevel access of memory
+        """
+        self._store=store
+
+    def __getitem__(self, key: int):
+        """
+        provide memory[offset] or memory[start:stop]
+        """
+        if self._store is None:
+            raise RuntimeError("you must call `set_store()` before using highlevel access")
+        return self.data_ptr(self._store)[key]
+
+    def __setitem__(self, key, value: bytearray):
+        """
+        provide setter for memory[key] with slice support
+
+        memory[offset]=value
+        memory[start:stop]=b'hello world'
+        memory[start:stop]=bytearray([1,2,3])
+        """
+        if self._store is None:
+            raise RuntimeError("you must call `set_store()` before using highlevel access")
+        data_ptr = self.data_ptr(self._store)
+        size = self.data_len(self._store)
+        if isinstance(key, int):
+            if key>=size:
+                raise IndexError("out of memory size")
+            return data_ptr[key]
+        if not isinstance(key, slice):
+            raise TypeError("key can either be integer index or slice")
+        start = key.start
+        stop = key.stop
+        if key.step is not None:
+            raise ValueError("slice with step is not supported")
+        if stop is None:
+            stop=start+len(value)
+        val_size = len(value)
+        if stop-start>val_size or stop>size:
+            raise IndexError("out of memory size")
+        src_ptr = (ctypes.c_ubyte * val_size).from_buffer(value)
+        dst_ptr = ctypes.addressof((ctypes.c_ubyte*val_size).from_address(ctypes.addressof(data_ptr.contents)+start))
+        ctypes.memmove(dst_ptr, src_ptr, val_size)
+
     def data_len(self, store: Storelike) -> int:
         """
         Returns the raw byte length of this memory.


wasmtime/_memory.py
```diff
@@ -62,6 +62,73 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
         """
         return ffi.wasmtime_memory_data(store._context, byref(self._memory))
 
+    def set_store(self, store: Storelike):
+        """
+        must be called to set the store for highlevel access of memory
+        """
+        self._store=store
```
alexcrichton, line 69: I think it would be better to not have this form of storage and take the store as input to the methods below. Additionally could they perhaps be renamed to `read` and `write` to be similar to the Wasmtime embedding and convey that copies are being performed as opposed to acquiring views?

wasmtime/_memory.py
```diff
@@ -62,6 +62,73 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+
+        start, stop, step = key.indices(size)
+        if step!=1:
+            raise ValueError("slice with step is not supported")
+
+        val_size = len(value)
+        if stop is None:
+            stop=start+val_size
+        if stop-start>val_size or stop>size:
+            raise IndexError("out of memory size")
```
alexcrichton, line 127: I think this logic is the same as above and here, so could this be refactored out?

wasmtime/_memory.py
```diff
@@ -62,6 +62,73 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+            raise ValueError("slice with step is not supported")
+
+        val_size = stop - start
+        value = bytearray(val_size)
+
+        if stop>size:
+            raise IndexError("out of memory size")
+        dst_ptr = (ctypes.c_ubyte * val_size).from_buffer(value)
+        src_ptr = ctypes.addressof((ctypes.c_ubyte*val_size).from_address(ctypes.addressof(data_ptr.contents)+start))
+        ctypes.memmove(dst_ptr, src_ptr, val_size)
```
alexcrichton, line 97: I think this may be able to be simplified a bit. There's a `to_bytes` helper in the `_ffi.py` file which I believe does roughly what this is doing. Could that helper function be used here, or perhaps the helper refactored a bit to be called with the pointer/len available on-hand here?

wasmtime/_memory.py
```diff
@@ -62,6 +62,73 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        if step!=1:
+            raise ValueError("slice with step is not supported")
+
+        val_size = len(value)
+        if stop is None:
+            stop=start+val_size
+        if stop-start>val_size or stop>size:
+            raise IndexError("out of memory size")
+        src_ptr = (ctypes.c_ubyte * val_size).from_buffer(value)
+        dst_ptr = ctypes.addressof((ctypes.c_ubyte*val_size).from_address(ctypes.addressof(data_ptr.contents)+start))
```
alexcrichton, line 129: The `ctypes.c_ubyte * val_size` type is "calculated" twice here, could that be refactored to happen only once?

alexcrichton (commented):
Thanks! Could you be sure to add some tests for this in the test suite, additionally for the cases that throw an error?

muayyad-alsadi:
> Thanks! Could you be sure to add some tests for this in the test suite, additionally for the cases that throw an error?

thank you for your quick response.
I hope this will get merged soon.

I have so far demonstrated two cases one is the [histogram matching image filter](https://medium.com/@alsadi/moving-hot-loops-from-python-to-wasm-wont-be-feasible-without-this-trick-65c9bd2dbe1b)
 the other is a simple hashing algorithm taken from the public domain constant CDB

https://github.com/muayyad-alsadi/wasm-demos/blob/main/cdb_djp_hash/

I believe the later one is a good use case for tests (the wasm is ~500 bytes only)

> I think it would be better to not have this form of storage and take the store as input to the methods below. Additionally could they perhaps be renamed to read and write to be similar to the Wasmtime embedding and convey that copies are being performed as opposed to acquiring views?

unfortunately we can't change the signature of `__getitem__()` and `__setitem__()` 
if you have a policy to decouple memory from storage, we can create a new class like this one I called it `FastMemory`

https://github.com/muayyad-alsadi/wasm-demos/blob/main/cdb_djp_hash/wasmtime_fast_memory.py

which is used like this

https://github.com/muayyad-alsadi/wasm-demos/blob/main/cdb_djp_hash/cdb_djp_hash.py

```python
store = Store()
module = Module.from_file(store.engine, 'cdb_djp_hash.wasm')
instance = Instance(store, module, [])
memory = instance.exports(store)['memory']
fast_mem = FastMemory(memory, store)
# ...
fast_mem[start:stop] = input
output = fast_mem[start_output:stop_output]
```

any suggestion for its name?

> I think this logic is the same as above and here, so could this be refactored out?

let me double check `start, stop, step = key.indices(size)` because it maybe handles all the edge conditions for us.

wasmtime/_memory.py
muayyad-alsadi: > The `ctypes.c_ubyte * val_size` type is "calculated" twice here, could that be refactored to happen only once?

I have an impression this is just a no-cost type casting, but let me review this and get back to you

```python
    #pr=Profile()
    #pr.enable()
    #for i in range(10000):
    #    cdb_djp_hash_wasm(large_a)
    #pr.disable()
    #pr.print_stats()
    #pr.print_stats('cumulative')
    #pr.print_stats('calls')
```

wasmtime/_memory.py
muayyad-alsadi: I've made an alternative implementation having which have common parts in  `_get_set_item_common()` 
Do you like this approach or do find it too complex?

https://github.com/muayyad-alsadi/wasm-demos/blob/main/cdb_djp_hash/wasmtime_fast_memory_alt.py

wasmtime/_memory.py
muayyad-alsadi: I made it once now

muayyad-alsadi:
I've added test, please let me know if you want me to move my logic into a different class and please suggest its name

wasmtime/_memory.py
muayyad-alsadi: yes, the logic of `to_bytes.()` worked fine, and it was fast. we just need away to pass custom offset and custom length

```python
def to_bytes(vec: wasm_byte_vec_t, offset=0, size=None) -> bytearray:
    if size is None: size=vec.size
    ty = c_uint8 * size
    return bytearray(ty.from_address(addressof(vec.data.contents)+offset))
```

muayyad-alsadi:
I don't know how do you want to solved those two type hints

```
91: error: Argument 1 to "len" has incompatible type "Union[int, bytearray]"; expected "Sized"  [arg-type]
86: error: Returning Any from function declared to return "Union[int, bytearray]"  [no-any-return]
```

should I change return to Any?

wasmtime/_memory.py
alexcrichton: Something like that is fine, yeah, or alternatively the logic inside of `to_bytes` could be split out to a separate function which takes a pointer/len argument and `to_bytes`, as-is right now, forwards to that.

wasmtime/_memory.py
alexcrichton: I think it would be good to deduplicate this logic, yes, since it's relatively tricky and important to get right.

alexcrichton:
> unfortunately we can't change the signature of

I realize this since they're Python builtins, which is why I was wondering if something like renaming the methods to `read`/`write` would work? I realize you wouldn't be able to use `[...]` syntax but that seems ok in the sense that ergonomics aren't of the utmost importance for something like this.

> if you have a policy to decouple memory from storage

Personally I think the "stateful" approach you have right now where `set_store` must be called first is a bit of a footgun. I think it would be better to express this through a different API. One alternative is what I was mentioning where `read` and `write` method names were used and one of their arguments is a `Storelike`. Another alternative is what you mentioned where, from a `Memory`, you can pair it with a `Store` and then read/write that returned object. I wouldn't call it a "fast memory" per se but instead probably something like:


```python
class Memory:
    def view(self, store: Storelike) -> MemoryView:
        # ...
```

where `MemoryView` has the `__{get,set}item__` methods you currently have, and `MemoryView` internally retains both the `Store` and `Memory`.

In either case though I think what's important is to avoid the stateful setting-of-a-store and instead make it explicit as part of the API.

> should I change return to Any?

Sorry I'm not familiar enough with mypy myself to know what's going on here.

muayyad-alsadi:
> where MemoryView

I've called it `MemorySlicer` because view might imply zero-copy.

alexcrichton:
Thinking more on this, can you say more as to why a dedicated `read`/`write` method won't work? Idiomatically with these bindings this new class doesn't really quite fit, since it's the only one that closes over a `Store` and everything else takes it as an argument.

muayyad-alsadi:
>  Thinking more on this, can you say more as to why a dedicated read/write method won't work? 

we can do them, but I have two concern

* the slicing operator is more familiar and versatile ex. `slicer[-10:]` or slicer[5:-10], where as taking arguments will be confusing. I'm even considering implementing step ex. `red, green, blue =slicer[0::3], slicer[1::3], slicer[2::3]`

https://github.com/muayyad-alsadi/wasm-demos/blob/main/cdb_djp_hash/cdb_djp_hash.py#L44
https://github.com/muayyad-alsadi/wasm-demos/blob/main/histogram-matching/histo_wasmtime_fast.py#L67

```python
    # cdb_djp_hash.py
    fast_mem[start:stop]=input

    # histo_wasmtime_fast.py
    fast_mem[0:] = a_ref
    fast_mem[len(a_ref):] = a_in
    dst = fast_mem[len(a_ref)+len(a_in): len(a_ref)+len(a_in)+len(a_in)]

```



*  I can't figure what the signature should be

```python
def read(self, store, key=None):
    start, stop, step = key.indices(size)

def write(self, store, value, key=None):
    start, stop, step = key.indices(size)
```

or

```python
def read(self, store, start=0, stop=None, step=None):
    key = slice(start, stop, step)
    start, stop, step = key.indices(size)
    # ...

def write(self, store, value, start=0, stop=None, step=None):
    key = slice(start, stop, step)
    start, stop, step = key.indices(size)
    # ...
```


* should we use built in `slice()` to get all the goodies of negative indexing ...etc or should we replicate that?
* should those function handle single offset access? in that case should it return an array of one item or just the item?
* is it ok for  `write()` to take `value` after `store`? while the standard `__setitem__()` is value-last

muayyad-alsadi:
after implementing read/write there are no exceptions to be handled, I'm know more into this solution.

muayyad-alsadi:
>  I'm even considering implementing step ex. `red, green, blue =slicer[0::3], slicer[1::3], slicer[2::3]`

I changed my mind, step is bad idea because it will do the expensive part 3 times

```python
# bad because logic is run 3 times
red, green, blue = mem.read(store, 0, None, 3), mem.read(store, 1, None, 3), mem.read(store, 2, None, 3)

# good
image = mem.read(store, 0) # expensive copying done once
red, green, blue = image[0::3], image[1::3], image[2::3], 
```

wasmtime/_memory.py
```diff
@@ -62,6 +64,57 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
         """
         return ffi.wasmtime_memory_data(store._context, byref(self._memory))
 
+    def read(
+            self,
+            store: Storelike,
+            start: typing.Optional[int] = 0,
+            stop: typing.Optional[int] = None) -> typing.Any:
```
alexcrichton, line 71: I may not have fully understood the mypy error from earlier, but I would natively expect this `typing.Any` to be `bytearray`. Does that run into issues though?

wasmtime/_memory.py
```diff
@@ -62,6 +64,57 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        """
+        provide fast way to read large memory slice similar to list[start:stop]
+        negative start, stop is allowed in a way similat to list slice mylist[-10:]
+        """
```
alexcrichton, line 75: Could this perhaps be renamed as:

```
Reads this memory starting from `start` and up to `stop` and returns a copy of the contents as a byte array.

The indexing behavior of this method is similar to `list[start:stop]` where negative starts can be used to read from the end, for example.
```

I think it's ok to drop "fast way" since it's the only way in the API to do this, and I think it's important to document that a copy is being performed.

wasmtime/_memory.py
```diff
@@ -62,6 +64,57 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        ptr_type = ctypes.c_ubyte * val_size
+        src_ptr = (ptr_type).from_address(ctypes.addressof(data_ptr.contents) + start)
+        return bytearray(src_ptr)
+
+    def write(
+            self,
+            store: Storelike,
+            value: typing.Any,
+            start: typing.Optional[int] = 0,
+            stop: typing.Optional[int] = None) -> typing.Any:
```
alexcrichton, line 93: Like above, could `value` and the return type here be typed as `bytearray`?

alexcrichton: Also, could `stop` be omitted here since it's implicitly the length of the `value` array?

wasmtime/_memory.py
```diff
@@ -62,6 +64,57 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        negative start, stop is allowed in a way similat to list slice mylist[-10:]
+        """
+        data_ptr = self.data_ptr(store)
+        size = self.data_len(store)
+        key = slice(start, stop, None)
+        start, stop, _ = key.indices(size)
+        val_size = stop - start
+        if val_size <= 0:
+            # return bytearray of size zero
+            return bytearray(0)
```
alexcrichton, line 83: Is this check necessary? (I'm not sure what `start`/`stop` combination would result in this)

wasmtime/_memory.py
```diff
@@ -62,6 +64,57 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        key = slice(start, stop)
+        start, stop, _ = key.indices(size)
+        # value must be bytearray ex. cast bytes() to bytearray
+        if not isinstance(value, array.array) and not isinstance(value, bytearray):
+            # value = array.array('B', value)
+            value = bytearray(value)
+        val_size = len(value)
+        # key.indices(size) knows about size but not val_size
+        stop = start + min(stop - start, val_size)
+        # NOTE: we can use * 1, because we need pointer to the start only
```
alexcrichton, line 109: I think this `NOTE` is outdated?

wasmtime/_memory.py
```diff
@@ -62,6 +64,57 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        ptr_type = ctypes.c_ubyte * val_size
+        src_ptr = (ptr_type).from_buffer(value)
+        dst_ptr = ctypes.addressof(
+            (ptr_type).from_address(ctypes.addressof(data_ptr.contents) + start)
+        )
```
alexcrichton, line 114: This does an addressof of a from_address expression, but is that necessary?

wasmtime/_memory.py
muayyad-alsadi: > Does that run into issues though?

you are correct, this was to handle it being `item[index]` which is single item not `bytearray`

wasmtime/_memory.py
muayyad-alsadi: > Also, could stop be omitted here since it's implicitly the length of the value array?

stop can be used to slice into value without specifying the start point
for example write last 10 bytes of value

```python
memory.write(store, value, None, -10)
memory.write(store, value, stop=-10)
```

wasmtime/_memory.py
muayyad-alsadi: yes, this is covered by the following test cases

```python
        # empty slices
        self.assertEqual(len(memory.read(store, 0, 0)), 0)
        self.assertEqual(len(memory.read(store, offset, offset)), 0)
        self.assertEqual(len(memory.read(store, offset, offset - 1)), 0)
        # out of bound access returns empty array similar to list slice
        self.assertEqual(len(memory.read(store, size_bytes + 1)), 0)
```

wasmtime/_memory.py
muayyad-alsadi: no, it's not, the following code will work totally fine, since the actual size is read by `memmove()` and the pointer is just the start

```python
        # NOTE: we can use * 1, because we need pointer to the start only
        # ptr_type = ctypes.c_ubyte * val_size
        ptr_type = ctypes.c_ubyte * 1
        src_ptr = (ptr_type).from_buffer(value)
        dst_ptr = (ptr_type).from_address(ctypes.addressof(data_ptr.contents) + start)
        ctypes.memmove(dst_ptr, src_ptr, val_size)
```

wasmtime/_memory.py
muayyad-alsadi: correct

muayyad-alsadi:
when I reverted to bytearray types I got this

```
_____________________________ tests/test_memory.py _____________________________
78: error: Incompatible types in assignment (expression has type "None", variable has type "bytearray")  [assignment]
_____________________________ wasmtime/_memory.py ______________________________
117: error: Returning Any from function declared to return "bytearray"  [no-any-return]
```

muayyad-alsadi:
> 117: error: Returning Any from function declared to return "bytearray"  [no-any-return]

this is in the case of `value = array.array('B', something)`
I'll return number of bytes

wasmtime/_memory.py
alexcrichton: The "NOTE" here disagrees with the code below. The NOTE makes me expect `c_ubyte * 1` when below it shows up as `* val_size`. I think that discrepancy should be fixed.

wasmtime/_memory.py
```diff
@@ -62,6 +64,64 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        return number of bytes written
+        """
+        data_ptr = self.data_ptr(store)
+        size = self.data_len(store)
+        key = slice(start, stop)
+        start, stop, _ = key.indices(size)
+        # value must be bytearray ex. cast bytes() to bytearray
+        if not isinstance(value, array.array) and not isinstance(value, bytearray):
+            # value = array.array('B', value)
+            value = bytearray(value)
```
alexcrichton, line 110: Since this is a relatively low-level interface could this perhaps take `bytearray` directly? I think that would help with mypy typing as well since the argument would have a specific type.

wasmtime/_memory.py
alexcrichton: The 0-length case seems like it can fall through to the below code just fine. I'm not familar with the negative-length case but that seems like it should throw an exception rather than silently returning an empty array?

wasmtime/_memory.py
```diff
@@ -62,6 +64,64 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        ptr_type = ctypes.c_ubyte * val_size
+        src_ptr = (ptr_type).from_address(ctypes.addressof(data_ptr.contents) + start)
+        return bytearray(src_ptr)
+
+    def write(
+            self,
+            store: Storelike,
+            value: typing.Any,
+            start: typing.Optional[int] = None,
+            stop: typing.Optional[int] = None) -> int:
```
alexcrichton, line 96: I think it would be best to remove the `stop` argument here. I understand you mentioned earlier that it can be used to reduce the size of the write but that can also be done by slicing the `value` argument here too. A downside of having the `stop` argument here is that this needs to document semantics of "what if the desired range is bigger than the `value` slice?" and I would prefer to not have to answer such a question.

alexcrichton: Additionally removing this `stop` argument removes the need to have `min`/etc logic and simplifies the indexing.

muayyad-alsadi:
TYT and I'm open to any suggestion or question.

wasmtime/_memory.py
j-blue-arz: I agree with @alexcrichton . Currently, the comment is somewhat confusing because it is not reflected in the code. It just educates the reader on something related to the code.
So I suggest to either remove the comment, or actually use the * 1 and explain in the comment why it works
I'd prefer the latter, e.g.:
```
# Recommended way to create an array type from data type: multiply with a positive integer
array_type = ctypes.c_ubyte * 1
...
```

wasmtime/_memory.py
j-blue-arz: @muayyad-alsadi wrote:
> I want this function to be used with b"123" and with array.array() without casting
because it's very common to pass bytes to bytearrays and vice versa
...
and bytearray and array.array are done without copying
the pointer as seen by ctypes.memmove is equivalent in both bases

I agree with that point. But creating a `bytearray` from `bytes` will involve a copy, right? Because `bytearray` is mutable. So maybe just hint at it in the documentation that supplying `bytes` will copy twice?

j-blue-arz:
Firstly, I think this is an important topic and I would like to help where I can.

I think much of the still-open discussions come from slice vs. no slice interface distinction. Both the question about the necessity of the `stop` parameter and the discussion on negative length can be reduced to that.

@muayyad-alsadi I understand that you would like the interface to behave like a slice. All your arguments on the above two matters are correct for slices. As a python programmer, I would also expect that `mem[start:stop] = val` works even with negative values for `stop`, and `mem[4:2]` will return an empty array.
However, it was already decided that we go with the `mem.write(store, start, stop, val)` interface. So it is not a slice! Now, negative `stop` values are somewhat confusing.
Even worse, your argument on the negative start / stop values does not hold when assigning to a slice:
```
>>> a = [0, 1, 2, 3, 4, 5, 6]
>>> a[2:-2] = [9]
>>> a
[0, 1, 9, 5, 6]
```
That is not at all what I would expect when writing to a location in memory, and also is (luckily) not what `memmove` does. Therefore, @alexcrichton's remark on the 
> semantics of "what if the desired range is bigger than the value slice?"
is completely appropriate.

I would vote to remove the `stop` parameter for the `write` method. For the `read` method, I am undecided.

wasmtime/_memory.py
muayyad-alsadi: @j-blue-arz I've just added such a comment.

wasmtime/_memory.py
muayyad-alsadi: @j-blue-arz I've removed the comment, and explanation why it worked because `ctypes.memmove(dst_ptr, src_ptr, val_size)` takes address of type `ctypes.c_char_p` or `ctypes.c_void_p` (it does not read size from the array size but rather from the 3rd param of `memmove`)

wasmtime/_memory.py
```diff
@@ -62,6 +64,67 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+            # return bytearray of size zero
+            return bytearray(0)
+        ptr_type = ctypes.c_ubyte * val_size
+        src_ptr = (ptr_type).from_address(ctypes.addressof(data_ptr.contents) + start)
+        return bytearray(src_ptr)
+
+    def write(
+            self,
+            store: Storelike,
+            value: typing.Any,
```
alexcrichton, line 94: When I said "works with mypy" better I was referring to the fact that this is listed as `typing.Any`. While I don't dispute "mypy runs without errors" falling back on `typing.Any` defeats the purpose of using type ascription in the first place.

This function does not take any type in Python. I think you provided a compelling case for it shouldn't just take `bytearray`, but in that case the list of supported types I think should be listed here under `typing.Union[...]` or something similar.

wasmtime/_memory.py
```diff
@@ -62,6 +64,67 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+            # value = array.array('B', value)
+            value = bytearray(value)
+        val_size = len(value)
+        if val_size == 0:
+            return val_size
+        # stop is exclusive
+        stop = start + val_size
+        slice_size = stop - start
+        if slice_size != val_size:
+            raise IndexError(f"mismatched value size, value of size [{val_size}] with slice of size [{slice_size}]")
```
alexcrichton, line 119: I don't think that this logic is correct? Here `slice_size` is always equal to `val_size`, so this `if` never fires?

wasmtime/_memory.py
```diff
@@ -62,6 +64,67 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        if stop > size:
+            raise IndexError("index out of range")
+        ptr_type = ctypes.c_ubyte * val_size
+        src_ptr = (ptr_type).from_buffer(value)
+        dst_ptr = (ptr_type).from_address(ctypes.addressof(data_ptr.contents) + start)
```
alexcrichton, line 124: I don't think that the parentheses around `ptr_type` are necessary here.

muayyad-alsadi:
I've addressed all of those points in my last push.

wasmtime/_memory.py
```diff
@@ -62,6 +64,66 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+        this happens when start offset is >= size or when end side of value is >= size
+        """
+        data_ptr = self.data_ptr(store)
+        size = self.data_len(store)
+        key = slice(start, None)
+        start = key.indices(size)[0]
+        if start >= size:
+            raise IndexError("index out of range")
+        # value must be bytearray ex. cast bytes() to bytearray
+        if not isinstance(value, array.array) and not isinstance(value, bytearray):
```
alexcrichton, line 111: I think that if `array.array` is to be supported here it needs more work and more tests. Currently there are no tests using `array.array` and for any size bigger than u8 I don't believe it's correct. For example an `array.array` of 32-bit integers doesn't correctly copy the size of the array into memory. Furthermore endianness is a concern with `array.array` where I don't think that larger than 1-byte values should be supported.

wasmtime/_memory.py
```diff
@@ -62,6 +64,66 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+            # value = array.array('B', value)
```
alexcrichton, line 112: I think this line can be removed?

wasmtime/_memory.py
```diff
@@ -62,6 +64,66 @@ def data_ptr(self, store: Storelike) -> "ctypes._Pointer[c_ubyte]":
...
+            value = bytearray(value)
```
alexcrichton, line 113: Is `bytearray` the best choice of conversion here? If the input is a `bytes`, for example, wouldn't this create a temporary copy before then copying into the linear memory?

wasmtime/_memory.py
muayyad-alsadi: you are totally right, should I check for `value.itemsize==1` and throw ValueError or remove `array.array`

muayyad-alsadi:
I would love to include [thewtex](https://github.com/thewtex) suggestion

```python
def get_buffer_ptr(self, store):
    """
    buffer suitable for creating zero-copy writable NumPy Buffer Protocol

    np_mem = np.frombuffer(memory.get_buffer_ptr(store), dtype=np.uint8)
    np_mem[start:end] = A
    B = np_mem[start:end]
    """
    ptr_type = ctypes.c_ubyte * (self.data_len(store))
    return ptr_type.from_address(ctypes.addressof(self.data_ptr(store).contents))
```

PR #138, opened 2023-03-24
Replace flake8 with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The `ruff` Action uses minimal steps to run in ~10 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image


alexcrichton (commented):
Thanks! Could you elaborate a bit more on why the switch though? This isn't a particularly large project and doesn't really stress flake8, but if it's more idiomatic nowadays to use ruff that seems reasonable to me.

Othrewise though some things I've noticed are:

* This adds a `pyproject.toml` which IIRC is intended to house many tooling configurations for a Python project. Do you know if some other configuration files, like `pytest.ini`, should go in this file? I don't know much about `pyproject.toml` myself but it seems like at least in spirit it's intneded to be the "one place" things are configured.
* This removes running the formatting tests during `pytest`, could that be added back in with ruff?

alexcrichton:
Sorry I don't really know what a "rule" is in this context and at the scale of this project (small) the speed doesn't matter too much. I'd ideally prefer to keep `pytest` running these tests since otherwise it seems more inconvenient to running locally? (e.g. it's an extra tool to execute during development)

Issue #96, opened 2022-09-22
Performance questions


alexprengere:
Incredible work! I was trying to see what kind of overhead there is to call a wasm function from Python.
I am using WSL2 on Windows, with a recent Fedora, CPython3.10.
Re-using the exact [gcd.wat](https://github.com/bytecodealliance/wasmtime-py/blob/main/examples/gcd.wat) from the examples, it looks like any call has a "cost" of about 30μs.

The code I used is a simple timer to compare performance:

```python
import wasmtime.loader
import time
from math import gcd as math_gcd
from gcd import gcd as wasm_gcd

def python_gcd(x, y):
    while y:
        x, y = y, x % y
    return abs(x)

N = 1_000
for gcdf in math_gcd, python_gcd, wasm_gcd:
    start_time = time.perf_counter()
    for _ in range(N):
        g = gcdf(16516842, 154654684)
    total_time = time.perf_counter() - start_time
    print(f"{total_time / N * 1_000_000:8.3f}μs")
```

This returns about:

```bash
   0.152μs  # C code from math.gcd
   0.752μs  # Python code from python_gcd
  31.389μs  # gcd.wat
```

Note that I tested this with an empty "hello world", and the 30μs are still there.
I am wondering about 2 things:
+ is this overhead inevitable and linked to the design, or are there ways to reduce it?
+ I noticed in the gcd.wat code that input parameters are expected to be i32, but the code does not fail when parameters exceed 2**32, so I was wondering how this works

alexcrichton:
Not much attention has been paid to performance in the bindings in this repository, so this isn't altogether unsurprising. I have never personally written high-performance Python and consequently probably made a ton of mistakes when writing these bindings.

If you're looking for performance though I would recommend the Rust bindings rather than the Python bindings, as there we have indeed focused on performance and the overhead is significantly smaller.

alexprengere:
Thanks, I was actually wondering if using wasm was a possibility for Python libraries that rely on C/C++/Rust, as you would only need to build the one file, then read it using wasmtime-py, instead of generating many platform wheels, which can be a headache. So I wanted to compare performance as a first step.

Thanks a lot for your response, I will close the issue.

alexcrichton:
That's definitely an intended use case for bindings like this, and I'll emphasize again that the cost here isn't intrinsic to these bindings or Python, it's probably just that I don't know how to write high-performance Python. PRs are of course always welcome for improvements as well.

[...]

PR #139, opened 2023-03-25
FIXES #137: make calling wasm from python 7x faster

### LINKED ISSUE #137: optimize time wasted in passing argument and fetching return
While discussing #96 I was able to identify a bottleneck
there was needlessly large time wasted in passing the arguments and retrieving the return

basically I made a simple 32-bit hash function [cdb_djp_hash.c](https://github.com/muayyad-alsadi/wasm-demos/blob/main/cdb_djp_hash/cdb_djp_hash.c)

in pure python the time is proportional to string length
in wasm it was almost constant time ~40ms regardless of string length 13 bytes to 1300 bytes (100x)
which means that most of those 40ms are in passing the arguments not in the actual loop (if it was the loop it would be proportional to string length or a large fraction of that).

This was confirmed using 

```python
pr=Profile()
pr.enable()
for i in range(10000):
        cdb_djp_hash_wasm(large_a)
pr.disable()
#pr.print_stats()
#pr.print_stats('cumulative')
pr.print_stats('calls')
```

the first note that the actual WASM call is fast (169ms out of 800ms)
which means that we might able to squeeze ~650ms of 800ms making it 4x faster

and I was able to identify bottleneck

- things that take too much cumulative time
- things that are called more than 10k (number of iterations)

examples of such things

- too many calls to `isinstance` which is 170k (in a 10k iteration benchmark), that is 17 times per call
- converting the argument `_value.py:129(_convert)` too ~200ms out of 800ms
- we got dynamically sized lists operations like `{method 'append' of 'list' objects}` 

Are we sure that passing arguments / converting can be made faster?
Yes

I can confirm `_value.py:129(_convert)`  that is too slow

I've moved the conversion outside the loop and confirmed it was the bottleneck

```python
    i32_start = Val.i32(start)
    i32_size = Val.i32(size)
    for i in range(10000):
        np_mem[start:stop]=input
        hash_wasm = ctypes.c_uint32(_cdb_djp_hash_wasm(i32_start, i32_size)).value
```

also I can confirm that this time is just waste because passing parameter should do the following

- convert little endian (which should not take time see below)
- set memory just after the stack pointer (which is now fast using #135 #81 #134 )
- move the stack pointer (which should not take time)

to proof that we can convert in almost no time

```
in_struct=Struct('<LL')
in_pack=in_struct.pack
for i in range(10000):
    b=pack(start, size)
```

muayyad-alsadi:
this is a great enhancement from 43 ms to 6 ms in @alexprengere question in #96 

```
0.0002523580042179674 # math_gcd
0.0014094869984546676 # python_gcd
0.043804362998344004  #  wasm_gcd
0.005873051006346941  # wasm_gcd_alt <-------- my change
```

### DIFF
--- a/wasmtime/_func.py
+++ b/wasmtime/_func.py
@@ -7,7 +7,7 @@ from ._extern import wrap_extern
 from typing import Callable, Optional, Generic, TypeVar, List, Union, Tuple, cast as cast_type, Sequence
 from ._exportable import AsExtern
 from ._store import Storelike
-
+from ._bindings import wasmtime_val_raw_t
 
 T = TypeVar('T')
 FUNCTIONS: "Slab[Tuple]"
@@ -27,11 +27,11 @@ class Func:
         set to `True` then the first argument given to `func` is an instance of
         type `Caller` below.
         """
-
         if not isinstance(store, Store):
             raise TypeError("expected a Store")
         if not isinstance(ty, FuncType):
             raise TypeError("expected a FuncType")
+        self._func_call_init(ty)
         idx = FUNCTIONS.allocate((func, ty.results, access_caller))
         _func = ffi.wasmtime_func_t()
         ffi.wasmtime_func_new(
@@ -56,6 +56,33 @@ class Func:
         ptr = ffi.wasmtime_func_type(store._context, byref(self._func))
         return FuncType._from_ptr(ptr, None)
 
+    def _func_call_init(self, ty):
+        self._ty = ty
+        ty_params = ty.params
+        ty_results = ty.results
+        self._params_str = (str(i) for i in ty_params)
+        self._results_str = (str(i) for i in ty_results)
+        params_n = len(ty_params)
+        results_n = len(ty_results)
+        self._params_n = params_n
+        self._results_n = results_n
+        n = max(params_n, results_n)
+        self._vals_raw_type = wasmtime_val_raw_t*n
+
+    def _create_raw_vals(self, *params: IntoVal) -> ctypes.Array[wasmtime_val_raw_t]:
+        raw = self._vals_raw_type()
+        for i, param_str in enumerate(self._params_str):
+            setattr(raw[i], param_str, params[i])
+        return raw
+    
+    def _extract_return(self, vals_raw: ctypes.Array[wasmtime_val_raw_t]) -> Union[IntoVal, Sequence[IntoVal], None]:
+        if self._results_n==0:
+            return None
+        if self._results_n==1:
+            return getattr(vals_raw[0], self._results_str[0])
+        # we can use tuple construct, but I'm using list for compatability
+        return [getattr(val_raw, ret_str) for val_raw, ret_str in zip(vals_raw, self._results_str)]
+
     def __call__(self, store: Storelike, *params: IntoVal) -> Union[IntoVal, Sequence[IntoVal], None]:
         """
         Calls this function with the given parameters
@@ -70,45 +97,30 @@ class Func:
         Note that you can also use the `__call__` method and invoke a `Func` as
         if it were a function directly.
         """
-
-        ty = self.type(store)
-        param_tys = ty.params
-        if len(params) > len(param_tys):
+        params_n = len(params)
+        if params_n > self._params_n:
             raise WasmtimeError("too many parameters provided: given %s, expected %s" %
-                                (len(params), len(param_tys)))
-        if len(params) < len(param_tys):
+                                (params_n, self._params_n))
+        if params_n < self._params_n:
             raise WasmtimeError("too few parameters provided: given %s, expected %s" %
-                                (len(params), len(param_tys)))
-
-        param_vals = [Val._convert(ty, params[i]) for i, ty in enumerate(param_tys)]
-        params_ptr = (ffi.wasmtime_val_t * len(params))()
-        for i, val in enumerate(param_vals):
-            params_ptr[i] = val._unwrap_raw()
-
-        result_tys = ty.results
-        results_ptr = (ffi.wasmtime_val_t * len(result_tys))()
-
+                                (params_n, self._params_n))
+        vals_raw = self._create_raw_vals(*params)
+        vals_raw_ptr = ctypes.cast(vals_raw, ctypes.POINTER(wasmtime_val_raw_t))
+        # according to https://docs.wasmtime.dev/c-api/func_8h.html#a3b54596199641a8647a7cd89f322966f
+        # it's safe to call wasmtime_func_call_unchecked because
+        # - we allocate enough space to hold all the parameters and all the results
+        # - we set proper types
+        # - but not sure about "Values such as externref and funcref are valid within the store being called"
         with enter_wasm(store) as trap:
-            error = ffi.wasmtime_func_call(
+            error = None
+            ffi.wasmtime_func_call_unchecked(
                 store._context,
                 byref(self._func),
-                params_ptr,
-                len(params),
-                results_ptr,
-                len(result_tys),
+                vals_raw_ptr,
                 trap)
             if error:
                 raise WasmtimeError._from_ptr(error)
-
-        results = []
-        for i in range(0, len(result_tys)):
-            results.append(Val(results_ptr[i]).value)
-        if len(results) == 0:
-            return None
-        elif len(results) == 1:
-            return results[0]
-        else:
-            return results
+        return self._extract_return(vals_raw)
 
     def _as_extern(self) -> ffi.wasmtime_extern_t:
         union = ffi.wasmtime_extern_union(func=self._func)


muayyad-alsadi:
Why I'm getting `'Func' object has no attribute '_params_n'`
despite it's being defined in the constructor `params_n = len(ty_params); self._params_n = params_n`

```
_______________________ ERROR collecting examples/gcd.py _______________________
examples/gcd.py:10: in <module>
    print("gcd(6, 27) = %d" % gcd(store, 6, 27))
wasmtime/_func.py:105: in __call__
    if params_n > self._params_n:
E   AttributeError: 'Func' object has no attribute '_params_n'
```

muayyad-alsadi:
the reason it failed was because it __init__ was never called as it was constructed with  `Func._from_raw()`
I would like to call `_init_call(ty)` but I don't have ty at that point.

muayyad-alsadi:
all the test errors are in `codegen`, what are so special about those? why normal work and this does not work?
can anybody take it from here?

wasmtime/_func.py
```diff
@@ -56,6 +63,36 @@ def type(self, store: Storelike) -> FuncType:
         ptr = ffi.wasmtime_func_type(store._context, byref(self._func))
         return FuncType._from_ptr(ptr, None)
 
+    def _create_raw_vals(self, *params: IntoVal) -> ctypes.Array[wasmtime_val_raw_t]:
+        raw = self._vals_raw_type()
+        for i, param_str in enumerate(self._params_str):
+            setattr(raw[i], param_str, params[i])
```
alexcrichton, line 69: I don't think that this conversion is quite correct. The `IntoVal` type could also be `Val`, for example, to handle things like v128/funcref/externref/etc. In such a case the raw attribute of `raw[i]` can't be set to that value, and I'm not sure what happens otherwise during such a process.

Additionally it's not safe to simply set the `externref` and `funcref` fields. Those need to go through the methods in the `Val` type.

alexcrichton: For an example you can see the logic in `Value._convert`

alexcrichton (commented):
This seems reasonable to me to add one way or another, but there's a number of soundness and safety issues to address before merging, in addition to the test failures.

wasmtime/_func.py
```diff
@@ -56,6 +63,36 @@ def type(self, store: Storelike) -> FuncType:
...
+        return raw
+    
+    def _extract_return(self, vals_raw: ctypes.Array[wasmtime_val_raw_t]) -> Union[IntoVal, Sequence[IntoVal], None]:
+        if self._results_n==0:
+            return None
+        if self._results_n==1:
+            return getattr(vals_raw[0], self._results_str0)
```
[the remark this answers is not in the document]
muayyad-alsadi: I'm trying to move all the logic in the constructor (or in `_init_call`) so that this logic is run once the module is loaded not on every single call. and hence the speed boost.

I need [a minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) example about the two missing types, in a way similar to my SIMD example.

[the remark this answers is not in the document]
muayyad-alsadi: I found the example it's `test_refs.py` I'll look into it and make sure it works

alexcrichton:
Sorry I don't have a ton of time to sink into this myself. I can help out more once CI is passing perhaps, but until then I think it's probably best to fix those errors. Once CI is passing that should give a sense of everything needed to handle various edge cases correctly.

muayyad-alsadi:
it was very difficult to know how `extenref` and `funcref` internally works
but I figured that out.

the code now works and passes all the tests except cosmetic tests
I'm not going to fix them right now, because I want to move some logic to `_value.py` and restructure the code related to `wasmtime_val_raw_t` and its helpers, this should not live in `_func.py`.

muayyad-alsadi:
there seems to be an existing typing bug (not introduced by my commits)

```python
_wasm_valtype_kind = dll.wasm_valtype_kind
_wasm_valtype_kind.restype = wasm_valkind_t
_wasm_valtype_kind.argtypes = [POINTER(wasm_valtype_t)]
def wasm_valtype_kind(arg0: Any) -> wasm_valkind_t:
    return _wasm_valtype_kind(arg0)  # type: ignore
```

the function `wasm_valtype_kind` returns `int` but is defined to return `wasm_valkind_t = c_uint8`
which is not compatible as c_uint8 is accessed via `ret.value`


```python
# _types.py
class ValType:
    def __str__(self) -> str:
        assert(self._ptr is not None)
        kind = ffi.wasm_valtype_kind(self._ptr)
        if kind == ffi.WASM_I32.value:
            return 'i32'
        if kind == ffi.WASM_I64.value:
            return 'i64'
        if kind == ffi.WASM_F32.value:
            return 'f32'
        if kind == ffi.WASM_F64.value:
            return 'f64'
        if kind == ffi.WASM_ANYREF.value:
            return 'anyref'
        if kind == ffi.WASM_FUNCREF.value: # <--- as you can see it's not kind.value but just kind which means it's int not c_uint8
            return 'funcref'
        return 'ValType(%d)' % kind  # type: ignore  <----------- this was kind.value which is wrong
```

If I change `wasm_valtype_kind` type to be int CI breaks with strange message
and it seems that this is a known issue because "type: ignore" was there (not added by me)

muayyad-alsadi:
I think I'm done and you can take it from here
since the typing issue is not introduced by me and all the CI passes.

muayyad-alsadi:
I've kept old method of calling so that we can compare the performance of the two `gcd._call_val(store, 6, 27)` and `gcd._call_raw(store, 6, 27)`

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
...
+
+    if attr == 'externref':
+        ptr = ctypes.POINTER(wasmtime_externref_t)
+        if not val:
+            return None
+        ffi = ptr.from_address(val)
+        if not ffi:
+            return None
+        extern_id = wasmtime_externref_data(ffi)
+        return _unintern(extern_id)
```
alexcrichton, line 55: Can this avoid duplicating `Val._as_externref`?

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
...
+    elif attr == 'funcref':
+        if val == 0:
+            return None
+        f = wasmtime_func_t()
+        f.store_id = store_id
+        f.index = val
+        return wasmtime.Func._from_raw(f)
```
alexcrichton, line 62: Can this avoid duplicating `Val._as_func`?

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
     return Val._id_to_extern.get(val)
 
 
+def get_valtype_attr(ty: ValType) -> str:
+    return val_id2attr[wasm_valtype_kind(ty._ptr)]  # type: ignore
```
alexcrichton, line 41: How come this is `type: ignore`?

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
     return Val._id_to_extern.get(val)
 
 
+def get_valtype_attr(ty: ValType) -> str:
```
alexcrichton, line 40: Can this have documentation for what it's doing?

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
...
+            casted = ctypes.addressof(val._raw.of.externref)
+        else:
+            ex = Val.externref(val)._raw
+            if ex:
+                casted = ctypes.addressof(ex.of.externref)
+            else:
+                casted = 0
+    elif attr == 'funcref':
+        if isinstance(val, Val) and val._raw and val._raw.kind == WASMTIME_FUNCREF.value:
+            casted = val._raw.of.funcref.index
```
alexcrichton, line 79: I don't think that this is quite right because it's not copying over the `store_id`. Same for below this is only copying in the `index` and not the `store_id`. Additionally the store needs to be checked for this branch.

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
...
+        elif isinstance(val, wasmtime.Func):
+            if val._func.store_id != store_id:
+                raise WasmtimeError("passed funcref does not belong to same store")
+            casted = val._func.index
+        else:
+            raise WasmtimeError("expecting param of type funcref got " + type(val).__name__)
+    else:
+        if isinstance(val, Val):
+            if val._raw:
```
alexcrichton, line 88: I think it would be best to `assert(val._raw)` here and a few places above. Handling a case where that is `None` I don't think is appropriate to use `casted = 0` in all situations.

wasmtime/_func.py
```diff
@@ -110,6 +136,52 @@ def __call__(self, store: Storelike, *params: IntoVal) -> Union[IntoVal, Sequenc
         else:
             return results
 
+    def _call_raw(self, store: Storelike, *params: IntoVal) -> Union[IntoVal, Sequence[IntoVal], None]:
+        """
+        internal implementation of calling a function that uses `wasmtime_func_call_unchecked`
+        """
+        if getattr(self, "_ty", None) is None:
```
alexcrichton, line 143: Instead of falling back to `getattr` can the mypy-listed-type of `self._ty` be `Optional`?

wasmtime/_func.py
```diff
@@ -16,6 +17,13 @@
 
 class Func:
     _func: ffi.wasmtime_func_t
+    _ty: FuncType
+    _params_n: int
+    _results_n: int
+    _params_str: list[str]
```
alexcrichton, line 23: Is it still performant to store strings here? I would expect naively that storing the integer kinds would be more efficient since then checking for funcref/externref would be integer comparisons instead of string comparisons.

wasmtime/_func.py
```diff
@@ -16,6 +17,13 @@
...
+    _results_str: list[str]
+    _results_str0: str
```
alexcrichton, line 25: I think this would be best stored as `Optional[str]` to avoid having a dummy value listed for empty results.

wasmtime/_func.py
```diff
@@ -56,21 +64,39 @@ def type(self, store: Storelike) -> FuncType:
...
+        return raw
 
-        Note that you can also use the `__call__` method and invoke a `Func` as
-        if it were a function directly.
+    def _extract_return(self, vals_raw: ctypes.Array[wasmtime_val_raw_t]) -> Union[IntoVal, Sequence[IntoVal], None]:
+        if self._results_n == 0:
+            return None
+        if self._results_n == 1:
+            return val_getter(self._func.store_id, vals_raw[0], self._results_str0)
+        # we can use tuple construct, but I'm using list for compatability
```
alexcrichton, line 78: This comment can be removed. Either the result should continue to be a list in which case there's no need for the comment, or the result should change to a tuple, in which case there's no need for the comment.

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
...
+                casted = ctypes.addressof(ex.of.externref)
+            else:
+                casted = 0
+    elif attr == 'funcref':
+        if isinstance(val, Val) and val._raw and val._raw.kind == WASMTIME_FUNCREF.value:
+            casted = val._raw.of.funcref.index
+        elif isinstance(val, wasmtime.Func):
+            if val._func.store_id != store_id:
+                raise WasmtimeError("passed funcref does not belong to same store")
+            casted = val._func.index
```
alexcrichton, line 83: Passing the `index` here is not correct. This needs to invoke `wasmtime_func_to_raw`. This also shows that this needs to have tests passing funcref, for example.

wasmtime/_value.py
```diff
@@ -25,6 +37,63 @@ def _unintern(val: int) -> typing.Any:
...
+
+def val_setter(store_id: int, dst: wasmtime_val_raw_t, attr: str, val: "IntoVal") -> None:
+    casted: typing.Union[Any, int, float, None, wasmtime.Func]
+    if attr == 'externref':
+        if isinstance(val, Val) and val._raw and val._raw.kind == WASMTIME_EXTERNREF.value:
+            casted = ctypes.addressof(val._raw.of.externref)
+        else:
+            ex = Val.externref(val)._raw
+            if ex:
+                casted = ctypes.addressof(ex.of.externref)
```
alexcrichton, line 74: Like below for funcref I don't think that this is correct. This needs to call `wasmtime_externref_to_raw`.

alexcrichton (commented):
I think there are still some correctness issues related to funcref/externref. The C API has `wasmtime_{func,externref}_{to,from}_raw` which need to be used for the various conversions and should avoid the need to pass around a `store_id` which should be an internal implementation detail.

Issue #141, opened 2023-04-10
Global.value should be a None check instead of a "truthy" check


thetianshuhuang:
The implementation of `Global.value` checks whether `val.value` is truthy instead of whether it is None.
- `Val.value` returns None if the value can't be represented by python according to the docstring, so a None check instead of a truthy check is probably the desired behavior
- The result of this is that if the global happens to have a value of 0, the Val object gets returned instead.

Current implementation is here:
```python
def value(self, store: Storelike) -> IntoVal:
    raw = ffi.wasmtime_val_t()
    ffi.wasmtime_global_get(store._context, byref(self._global), byref(raw))
    val = Val(raw)
    if val.value:
        return val.value
    else:
        return val
```

Suggested fix:
```python
def value(self, store: Storelike) -> IntoVal:
    raw = ffi.wasmtime_val_t()
    ffi.wasmtime_global_get(store._context, byref(self._global), byref(raw))
    val = Val(raw)
    if val.value is not None:
        return val.value
    else:
        return val
```

alexcrichton:
Looks reasonable to me! Would you be up to send a PR for this?

PR #142, opened 2023-04-10
Fix Global.value to use None check instead of truthy check

thetianshuhuang:
Implements the change suggested in #141.


alexcrichton:
Thanks! Would you be up for adding a test for this as well?

thetianshuhuang:
I added a test case in `TestGlobal` which creates a global with a falsy value (i64: 0), fetches, and asserts that the type is still int; hopefully this is the right place to put it.

alexcrichton:
Thanks!

PR #149, opened 2023-05-11
avoid ambiguous generated types

dicej:
While developing and testing `componentize-py`, I've sought to mimic `wasmtime-py`'s type binding generator.  In doing so, I noticed that certain types were ambiguous, e.g.:

```python
Option[Option[int]] # None is ambiguous

Foo = Union[int, float] # OK
Bar = Union[Foo, int, float] # e.g. 42 and 0.42 are ambiguous

Baz = Union[Optional[int], int] # e.g. 42 is ambiguous
```

This fixes the above cases by:
- Introducing a `Some` type for use with nested `option` types
- Disallowing nested raw `union`s and `option`s inside raw `union`s

### DIFF
--- a/rust/bindgen/src/bindgen.rs
+++ b/rust/bindgen/src/bindgen.rs
@@ -164,6 +164,24 @@ impl WasmtimePy {
         }
     }
 
+    fn print_some(&mut self) {
+        if !self.all_intrinsics.insert("some_type") {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
Thanks! Would it be possible to add a test for this as well?

alexcrichton (approved):

Issue #155, opened 2023-05-23
NameError: name 'RuntimeValueString' is not defined


danbev:
I ran into the following error when trying to extract types into a separate .wit files:
```console
Traceback (most recent call last):
  File "/home/danielbevenius/work/wasm/wasmtime-py/demo-module/demo.py", line 12, in <module>
    main()
  File "/home/danielbevenius/work/wasm/wasmtime-py/demo-module/demo.py", line 8, in main
    ret = wit.something(store)
          ^^^^^^^^^^^^^^^^^^^^
  File "/home/danielbevenius/work/wasm/wasmtime-py/demo-module/dist/exports/wit.py", line 27, in something
    variant = RuntimeValueString(list)
              ^^^^^^^^^^^^^^^^^^
NameError: name 'RuntimeValueString' is not defined

```
I've created a [reproducer] for this might help troubleshoot this and the readme contains a note about this perhaps being an issue with a missing module/namespace. The same .wit files work with JavaScript and Rust.

[reproducer]: https://github.com/danbev/wasmtime-py/blob/name-error/demo-module/README.md

alexcrichton:
Thanks for the report! This is probably a mismatch in how variants are represented and how they're lifted, and probably shouldn't be too hard to fix. I don't have time right now myself to tackle this but I'd be happy to review a PR!

PR #156, opened 2023-05-24
Add module name to missing assignment

danbev:
This commit adds the qualified module name to assignement that are imported from a separate modules.

The motivation for this is that currently this will cause a NameError. For example, if we have types defined in a Python module named demo_types, the variant assignement generated by bindgen will be:
```python
from ..imports import demo_types
variant = RuntimeValueString(list)
```
Which will generate the following error:
```console
  File "/wasmtime-py/demo-module/demo.py", line 12, in <module>
    main()
  File "wasmtime-py/demo-module/demo.py", line 8, in main
    ret = wit.something(store)
          ^^^^^^^^^^^^^^^^^^^^
  File "wasmtime-py/demo-module/dist/exports/wit.py", line 27,
  in something
    variant = RuntimeValueString(list)
              ^^^^^^^^^^^^^^^^^^
NameError: name 'RuntimeValueString' is not defined
```

With this commit, the generated assignement will instead be:
```python
from ..imports import demo_types
variant = demo_types.RuntimeValueString(list)
```

Fixes: https://github.com/bytecodealliance/wasmtime-py/issues/155

### DIFF
--- a/rust/bindgen/src/bindgen.rs
+++ b/rust/bindgen/src/bindgen.rs
@@ -1308,6 +1308,24 @@ impl InterfaceGenerator<'_> {
         self.src.pyimport(&module, iface.as_str());
         format!("{iface}.{ty}")
     }
+
+    fn fqn_name(&mut self, name: &str, type_id: &TypeId) -> String {
+        let ty = &self.resolve.types[*type_id];
+        if let Some(name) = &ty.name {
+            let owner = match ty.owner {
+                TypeOwner::Interface(id) => id,
+                _ => unreachable!(),
+            };
+            if Some(owner) != self.interface {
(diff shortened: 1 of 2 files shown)


danbev:
> Would you be up for adding a test for this?

Yeah, I'll try adding a test for this tomorrow. I've only been testing this against the reproducer, and against our component as I'm still not very familiar with the code base yet not exactly sure where a test for this would go. Should I add this to one of the tests in `tests` or should this be a new test?

alexcrichton:
Yeah [`tests/codegen`](https://github.com/bytecodealliance/wasmtime-py/tree/main/tests/codegen) would be where this would live

alexcrichton:
Thanks!

PR #171, opened 2023-08-14
escape keywords in generated bindings

### LINKED ISSUE #150: Avoid using Python keywords as identifiers in `bindgen`
Currently, the binding generator will use keywords like `in` when generating identifiers, leading to syntax errors at runtime.  It should e.g. suffix them with underscores instead.

https://docs.python.org/3/reference/lexical_analysis.html#keywords

`componentize-py` has the same problem, which is how I noticed this.  I'll look at fixing it both here and there.

dicej:
This avoids runtime syntax errors by adding an underscore suffix to identifiers which would otherwise match Python keywords.

Fixes #150


alexcrichton:
Thanks!

PR #176, opened 2023-08-22
support bare function guest imports

dicej:
This adds support for exporting bare (i.e. not contained in an interface) functions to the guest.  It works by generating a `Host` class (and corresponding `RootImports` field) containing each host function declaration.

### DIFF
--- a/rust/bindgen/src/imports.rs
+++ b/rust/bindgen/src/imports.rs
@@ -19,28 +19,6 @@ impl PyImports {
         push(&mut self.typing_imports, module, name.into())
     }
 
-    pub fn merge(&mut self, imports: PyImports) {
-        for (module, names) in imports.pyimports {
-            if let Some(names) = names {
-                for name in names {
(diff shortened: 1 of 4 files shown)


rust/bindgen/src/imports.rs
```diff
@@ -19,6 +19,28 @@ impl PyImports {
         push(&mut self.typing_imports, module, name.into())
     }
 
+    pub fn merge(&mut self, imports: PyImports) {
```
alexcrichton, line 22: I originally tried to avoid adding this method because imports are automatically renamed if they conflict with things and the merging process doesn't work well where something that wasn't renamed could be renamed otherwise during generation to make things work. 

Would it be possible to swap the `Source` objects perhaps during the generation of the root functions where the "root `Source`" temporarily becomes the generator `Source`? I forget if I've done that here in this bindings generator or only in others.

alexcrichton (approved):
Thanks!

alexcrichton (approved):

PR #184, opened 2023-09-08
ENH Support PathLike in `config.stdxxx_file = path`

hoodmane:
Generally Python APIs that take paths accept the path as a `str`, `bytes` or `PathLike`. This makes a few wasmtime-py apis follow this convention.


alexcrichton:
Thanks!

PR #185, opened 2023-09-08
MAINT Use `subprocess.run` rather than `os.popen`

hoodmane:
This is a very minor change, just making the code more idiomatic


alexcrichton:
Thanks!

PR #186, opened 2023-09-08
FIX Make it possible to convert a Python int/float to externref

hoodmane:
Before this PR the added test fails with:
```
TypeError: don't know how to convert 5 to anyref
```


alexcrichton:
Thanks!

PR #187, opened 2023-09-09
API Make InstanceExports into a Mapping

hoodmane:
This is a breaking API change. I expected `InstanceExports` to be a subclass of `Mapping`. This PR includes a few changes:

1. `__iter__` should return an iterator over the **keys** of the map, not over the values.
2. we are expected to implement `__getitem__`, a `.get` method is derived automatically.
3. It's a bit unexpected that you can index both by the keys and by the index. It makes it hard to satisfy mypy, and it makes it hard to satisfy the invariants of `Mapping` (for instance, that `val in map` iff `val in list(map)`). I think it's reasonably natural to expose directly a tuple, typed as a `Sequence`, with the exports if we want to index them. I called this field `by_index`. For good measure, I made it a getter so that attempts to assign to it fail.

### DIFF
--- a/tests/test_instance.py
+++ b/tests/test_instance.py
@@ -14,7 +14,7 @@ class TestInstance(unittest.TestCase):
         module = Module(store.engine, '(module (func (export "")))')
         instance = Instance(store, module, [])
         self.assertEqual(len(instance.exports(store)), 1)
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
Thanks!

Issue #189, opened 2023-09-21
wasmtime-py 13.0.0, 13.0.1 breaks compatibility with Python 3.7, 3.8, 3.9


whitequark:
See these failing jobs: https://github.com/YoWASP/runtime/actions/runs/6255807281/job/16985710642

Not sure if this was intended? README still advertises compatibility with Python 3.6.

thewtex:
Proposed patch in #190 

> wasmtime-py 14 

I think wasmtime-py 13 was meant.

alexcrichton:
Ok Wasmtime 13.0.1 is now published, but let me know if that's still broken!

whitequark:
I've also updated the title to be a bit more precise.

alexcrichton:
I'm no Python expert myself so you can see me flailing around on https://github.com/bytecodealliance/wasmtime-py/pull/192 trying to get something working. Right now I'm at the point where `Protocol` doesn't exist in 3.7. I'm not sure if there are more issues but it's as far as I've gotten.

`Protocol` is used by generated bindings for the component model, so it wasn't broken in the last release or anything like that. I think I may have fixed the issues that broke compat in the last release. Does anyone know how to only run a subset of tests on Python 3.8+? Ideally the `codegen` tests would all be disabled in 3.7 to sidestep the `Protocol` issue.

Apparently GitHub actions doesn't have Python 3.6 for the Ubuntu 22.04 image so I chose 3.7 as the minimum version.

whitequark:
> Does anyone know how to only run a subset of tests on Python 3.8+? Ideally the `codegen` tests would all be disabled in 3.7 to sidestep the `Protocol` issue.

[This maybe?](https://stackoverflow.com/a/61869801/254415)

> Apparently GitHub actions doesn't have Python 3.6 for the Ubuntu 22.04 image so I chose 3.7 as the minimum version.

No problem for me; I'll bump up the YoWASP runtime version then.

alexcrichton:
Hm I realize now that this is not only in tests but also the `wasmtime.bindgen` module which I don't necessarily know how to prevent `pytest` from loading for just 3.7. Given [this run though](https://github.com/bytecodealliance/wasmtime-py/pull/192) it looks like 3.8 passes tests.

@whitequark is it a problem if the minimum version is raised to 3.8?

whitequark:
Thanks! I've published an updated YoWASP runtime, it passes tests fine.

PR #190, opened 2023-09-21
FIX Python 3.7, 3.8 Union support

thewtex:
For #189


alexcrichton:
Thanks!

igrr:
It seems there is another instance here:
https://github.com/bytecodealliance/wasmtime-py/blob/8bdd1c41ef5e64f340409b3211346547f2bda544/wasmtime/_wasi.py#L10

Issue #204, opened 2024-01-23
Can't run module with tail calls


womeier:
As the title suggests, I'd like to run a Wasm module with tail calls.
Other proposals can be configured in the `Config` class, but not tail calls.

Would greatly appreciate it if there was an option for that as well.

alexcrichton:
Thanks for the report! Wasmtime supports the underlying feature here but the config option hasn't been surfaced yet from the Wasmtime C API into these bindings. If you're up for it a PR to add a method to expose this would be greatly appreciated! It'd look quite similar to the other methods of exposing wasm features.

womeier:
thanks!

Created a PR, but need help with testing it, see the PR.

PR #205, opened 2024-01-24
support configuring tail calls

womeier:
This PR adds support for configuring tail calls (Closes https://github.com/bytecodealliance/wasmtime-py/issues/204),

it depends on the change to the C api here: https://github.com/bytecodealliance/wasmtime/pull/7811,

without the change, the CI fails as expected with: `[...] _libwasmtime.so: undefined symbol: wasmtime_config_wasm_tail_call_set.`

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -9,6 +9,7 @@ class TestConfig(unittest.TestCase):
         config = Config()
         config.debug_info = True
         config.wasm_threads = True
+        config.wasm_tail_call = True
         config.wasm_reference_types = True
         config.wasm_simd = True
         config.wasm_bulk_memory = True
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! In terms of timing for this, is this ok for you to delay for a month-or-so until the Wasmtime 18 release? Wasmtime 17 will be released in a few days but won't have https://github.com/bytecodealliance/wasmtime/pull/7811, that's slated for the Wasmtime 18 release on merge which will be released on Feb 20.

alexcrichton:
I've rebased this and fixed a conflict after the [Wasmtime 18 update](https://github.com/bytecodealliance/wasmtime-py/pull/210) and I'll merge this for that release, thanks again for this!

PR #209, opened 2024-02-15
Enable explicit destruction of all resources

alexcrichton:
This commit adds a suite of methods to most of the types in this package to enable explicit destruction of resources as necessary. Specifically most classes which own a pointer internally now sport some new methods:

* `__enter__` and `__exit__` enable using the objects in `with` blocks.
* `close` enables explicitly deallocating the object.

All objects internally now validate that they have not been closed before operating, raising a `ValueError` if a closed object is operated on.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,4 +1,5 @@
 import unittest
+from contextlib import closing
 
 from wasmtime import *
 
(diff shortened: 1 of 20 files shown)


alexcrichton:
@dicej would you be up for reviewing this?

Also, as context, [this is the equivalent issue](https://github.com/bytecodealliance/wasmtime-go/issues/209) for the Go bindings of Wasmtime.

dicej (approved):
LGTM.  Thanks for doing this!

PR #215, opened 2024-03-21
Docker compose setup

sarunasazna2:
I tried to play with test locally and... Setup is kind of complicated, did not have rust installed, maybe lack of experience with rust tools (I am a python developer first), but - it did not work on my maschine :( Therefore created some rudimentary docker-compose/make setup (instalation steps more or less copied from github actions - minus the cache). Which - works on my machine (windows 11 with WSL2), should be all fine with Linux too.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,6 +46,24 @@ After you've completed the set up steps, you can run the tests locally with
 $ pytest
 ```
 
+### Docker-compose setup
+Docker installation guide can be found [here](https://docs.docker.com/engine/install/)
+In short - all standard docker-compose commands can be used, but we also have some as shortcuts in [Makefile](Makefile).
+```bash
+make start
+```
+To run hello world example:
+```bash
+make hello-world
+```
+To run tests:
+```bash
+make test
+```
+To stop the image:
+```bash
+make stop
+```
 ### CI and Releases
 
 The CI for this project does a few different things:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM python:3.11
+
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl
+# Get Rust
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+RUN curl https://wasmtime.dev/install.sh -sSf | bash
+
+COPY . /code/wasmtime-py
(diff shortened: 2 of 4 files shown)


alexcrichton:
Thanks for the PR! I'm personally a bit hesitant to add this sort of automation without any testing of it. I'm also wary of `Makefile`s myself because they tend to not be too friendly to Windows developers. 

Could this perhaps be tackled from a different angle where a `*.wasm` file is checked in so `build-rust.py` isn't necessary to start developing? (or something like that)

sarunasazna2:
Not saying that is the only way. The PR is more of a proposal to somehow solve "works on my machine problem".

Makefile is does not help with the "works on my machine thing" only it makes commands shorter.
However, because I am aware with the problems with Makefile and Windows  I did entire setup on Windows  Laptop (windows 11) and ran Makefile via Powershell. And just tested with CMD:
image
Cannot point exact to exact installation of makefile on this particular laptop, since I did a while ago. But usually I tend to go the easiest way possible. Most likelly used some command from [SO](https://stackoverflow.com/questions/2532234/how-to-run-a-makefile-in-windows). Not against moving it to python script or just providing raw `docker-compose` commands - they are in this case quite easy to remember :) And remembering is particulary important for windows - powershell is getting better, but the history is still more or less non existing (ctrl+R seems to be working, but it has a long way to go to reach even stock ubuntu bash level). What would be actually nice is to reuse the same tool locally as on CI. So basically putting distro related logic to the .py setup scripts themselves. What I did here at the end was copy-pasting CI logic.

Getting back to "works on my machine". At first I tried to setup the project locally, but installing rust on windows appeared to be too combursum. Might be my lack of experience/patience, but still - would be nice to have easier setup even without knowing much about Rust since the project has to convince Python developers to use it(my subjective opinion), and we are accustomed to easy setups :D

alexcrichton:
No I agree with much of what you say, I don't think that this should necessarily require Rust. My point was moreso I'd prefer to lift the need for needing this sort of configuration entirely. For example if the project is confusing enough to warrant a docker-compose.yml file I think it's best to fix the confusion rather than adding the configuration. This isn't intended to be a complicated Python package, the main weird part is the Rust support for component compiled to wasm.

Put another way, I think and alternative source to solving the difficulties you're having would be to check in a version of the Rust wasm binary into this repository (along with generated code it generates). That way it's not necessary to use `build-rust.py` during development, meaning that the Rust bits and container bits aren't needed.

alexcrichton:
I'm going to close this as I'd prefer to avoid extra setup in this repository.

PR #222, opened 2024-04-03
loader: assorted fixes

whitequark:
Depends on https://github.com/bytecodealliance/wasmtime-py/pull/221.

This is an incremental but substantial improvement, to the point where I'd say I consider this implementation correct, or very close to correct, according to my understanding of the importlib hooks involved.

It is however still not thread-safe, I will likely address that after adding component support.

### DIFF
--- a/wasmtime/_types.py
+++ b/wasmtime/_types.py
@@ -381,7 +381,7 @@ class ImportType(Managed["ctypes._Pointer[ffi.wasm_importtype_t]"]):
     @property
     def name(self) -> Optional[str]:
         """
-        Returns the name in the modulethis import type refers toa.
+        Returns the name in the module this import type refers to
 
         Note that `None` may be returned for the module linking proposal where
         the field name is optional.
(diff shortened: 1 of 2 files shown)


whitequark:
@alexcrichton I have a prototype of an importlib loader for components that builds on top of these fixes but I'll probably hold off submitting it until you've had a chance to review #221 and #222.

whitequark:
Rebased.

alexcrichton:
Thank you very much for this! All looks great to me and I'm glad this has a better pair of eyes on it than my own "not very Python trained" eyes.

whitequark:
> I'm glad this has a better pair of eyes on it than my own "not very Python trained" eyes.

Honestly, this particular bit of the Python library has almost nobody trained in it, and it's definitely one of the most advanced of advanced topics any Python programmer will ever explore, save for perhaps the more obscure FFI-related ones. It was really quite difficult for me to get my head around it, partly due to how obscure it is, partly due to the peculiar state of the docs, and I've spent basically a workday or two reading sources of the Python internals and comparing them across versions.

alexcrichton:
IIRC the initial bits here were copy/pasted from assorted blogs/stack overflow posts, so I suspect you've done much more studying already than I initially did!

Issue #223, opened 2024-04-04
bug: sched_yield is not defined


HarikrishnanBalagopal:
```
$ python3 main.py 
Traceback (most recent call last):
  File "............../test-py-wasmtime/main.py", line 63, in <module>
    main()
  File "............../test-py-wasmtime/main.py", line 53, in main
    instance = linker.instantiate(store=store, module=module)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/haribala/Documents/code/remote/github.com/HarikrishnanBalagopal/test-py-wasmtime/.venv/lib/python3.11/site-packages/wasmtime/_linker.py", line 171, in instantiate
    raise WasmtimeError._from_ptr(error)
wasmtime._error.WasmtimeError: unknown import: `wasi_snapshot_preview1::sched_yield` has not been defined
```

alexcrichton:
Thanks for the report! As I think you've found this needs a `linker.define_wasi()` call. Agreed the docs/examples could be better!

PR #224, opened 2024-04-04
loader: add support for components

### LINKED ISSUE #219: Backward compatibility contract for bindings generated by `wasmtime.bindgen`?
What is the compatibility contract for bindings generated by `wasmtime.bindgen`? Put differently, can I put the generated files into a wheel file or not?

If bindings generated with wasmtime-py X.0.0 are going to work on wasmtime-py (X+1).0.0: I can, and the deployment process of downstream code will consist of pre-generating bindings on CI and then distributing the wheel with a `wasmtime>=X` version constraint.

If bindings generated with wasmtime-py X.0.0 will only work on wasmtime-py X.0.0: I can not, and I have several options, all of which are very unpleasant:
1. I can distribute the wheel with a `wasmtime==X` version constraint. This is obviously correct, but if this wheel needs to be co-installable with anything else that uses wasmtime, it is not possible to use the [track-pypi-dependency-version](https://github.com/bytecodealliance/wasmtime-py?tab=readme-ov-file#versioning) script the README currently recommends, because it means that upgrading a completely unrelated package that bumps its lower bound on `wasmtime` may force an upgrade of *every* bindgen-using package. This is going to rapidly become impractical once you have, potentially, as few as two bindgen-using packages from different publishers.
2. I can distribute the wheel with a `wasmtime>=Y,wasmtime<X+1` version constraint as recommended in the README currently, only include the Wasm component in it, run `wasmtime.bindgen` on first use, and cache the generated files. This is very onerous on the consumer (every consumer would have to come up with a bespoke caching scheme; it's tempting to try and write to the site-packages where your own package is installed but it may not be wirtable; keeping the caching code in sync is burdensome in itself even if you factor it out in a separate package).
3. I can distribute the wheel with a `wasmtime>=Y,wasmtime<X+1` version constraint, and generate bindings for each wasmtime version from Y to X inclusive. This will be fairly painful to do on CI (in practice it requires a single virtual environment per wasmtime version, potentially a lockfile per wasmtime version, and some custom tooling to collect all the generated files) and result in rapid binary bloat unless carefully managed (at 14M core module size it won't be long until this wrapper for a 300 kB JS library will become the biggest PyPI package installed on the system).
4. I can distribute the wheel with a `wasmtime>0` version constraint. I mean, I personally would, but I think most consumers of the `wasmtime.bindgen` interface have neither time nor inclination to build complicated caching and versioning schemes, and they will simply accept a status quo of stochastic runtime breakage after wasmtime releases. I think this is doing a disservice to the ecosystem.

### LINKED ISSUE #220: `wasmtime.bindgen` shouldn't be using `__file__`
There's actually no guarantee at all that Python source files exist anywhere on the filesystem, and even if they do they might as well be inside a [zip archive](https://docs.python.org/3/library/zipimport.html) or even a [Rust executable](https://pyoxidizer.readthedocs.io/en/stable/index.html). Because of this using `__file__` for anything but debugging or logging, and in particular loading resources, is a fragile practice and packages like `wasmtime-py` should avoid it.

In practical terms the header of the generated code that looks like this:

```py
from .intrinsics import _decode_utf8, _encode_utf8, _load
import ctypes
import pathlib
import wasmtime

class Root:

    def __init__(self, store: wasmtime.Store) -> None:
        path = pathlib.Path(__file__).parent / ('root.core0.wasm')
        module = wasmtime.Module.from_file(store.engine, path)
```

should be replaced with:

```py
from .intrinsics import _decode_utf8, _encode_utf8, _load
import ctypes
import pathlib
import wasmtime
try:
    import importlib.resources as importlib_resources
    importlib_resources.files() # check if implicit anchor is supported
except TypeError:
    import importlib_resources

class Root:

    def __init__(self, store: wasmtime.Store) -> None:
        file = importlib_resources.files() / ('root.core0.wasm')
        module = wasmtime.Module.from_file(store.engine, file.read_bytes())
```

Note that the `importlib.resources.files()` (the zero argument form) is very new, [freshly added in 3.12](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files), and unfortunately consumers of the bindings will have to add the [`importlib-resources>=5.10`](https://pypi.org/project/importlib-resources/) polyfill as a dependency to support all versions of Python.

(Potentially you could have another fallback that uses `__file__` if neither the zero argument form is supported nor the polyfill is installed, but I find a three-deep chain of compatibility shims pretty hard to maintain.)

whitequark:
Fixes #220.
Closes #219.
Depends on #221.
Depends on #222.

Works like this:

```
$ ls
wavedrom.wasm    ...
$ python -c 'import wasmtime.loader, json, wavedrom; store = wasmtime.Store(); print(wavedrom.Root(store).render_json(store, json.dumps({"signal": [{ "name": "clk", "wave": "p..." }, { "name": "data", "wave": "0110" }]})))'
<svg xmlns="http://www.w3.org/2000/svg" ....
```

### DIFF
--- a/rust/src/bindgen.rs
+++ b/rust/src/bindgen.rs
@@ -539,16 +539,27 @@ impl<'a> Instantiator<'a> {
         let i = self.instances.push(idx);
         let core_file_name = self.gen.core_file_name(&self.name, idx.as_u32());
         self.gen.init.pyimport("pathlib", None);
+        self.gen.init.pyimport("importlib_resources", None);
 
         uwriteln!(
             self.gen.init,
-            "path = pathlib.Path(__file__).parent / ('{}')",
+            "file = importlib_resources.files() / ('{}')",
             core_file_name,
         );
+        uwriteln!(self.gen.init, "if isinstance(file, pathlib.Path):");
+        self.gen.init.indent();
         uwriteln!(
             self.gen.init,
-            "module = wasmtime.Module.from_file(store.engine, path)"
+            "module = wasmtime.Module.from_file(store.engine, file)"
         );
+        self.gen.init.dedent();
+        uwriteln!(self.gen.init, "else:");
+        self.gen.init.indent();
+        uwriteln!(
+            self.gen.init,
+            "module = wasmtime.Module(store.engine, file.read_bytes())"
+        );
+        self.gen.init.dedent();
         uwrite!(
             self.gen.init,
             "instance{} = wasmtime.Instance(store, module, [",
--- a/examples/loader_component.py
+++ b/examples/loader_component.py
@@ -1,0 +1,11 @@
+# This example shows how you can use the `wasmtime.loader` module to load wasm
+# components without having to generate bindings manually
+
+import wasmtime, wasmtime.loader
+
+import loader_component_add  # type: ignore
+
+
+store = wasmtime.Store()
+component = loader_component_add.Root(store)
+assert component.add(store, 1, 2) == 3
--- a/examples/loader_component_add.wat
+++ b/examples/loader_component_add.wat
@@ -1,0 +1,12 @@
+(component
+  (core module $C
+    (func (export "add") (param i32 i32) (result i32)
+      local.get 0
+      local.get 1
+      i32.add)
+  )
+  (core instance $c (instantiate $C))
+  (core func $add (alias core export $c "add"))
+  (func (export "add") (param "x" s32) (param "y" s32) (result s32)
+    (canon lift (core func $add)))
+)
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ setuptools.setup(
         "Source Code": "https://github.com/bytecodealliance/wasmtime-py",
     },
     packages=['wasmtime'],
+    install_requires=['importlib_resources>=5.10'],
     include_package_data=True,
     package_data={"wasmtime": ["py.typed"]},
     python_requires='>=3.8',
--- a/wasmtime/loader.py
+++ b/wasmtime/loader.py
@@ -7,14 +7,19 @@ can `import your_wasm_file` which will automatically compile and instantiate
 `your_wasm_file.wasm` and hook it up into Python's module system.
 """
 
+from typing import NoReturn, Iterator, Mapping, Dict
+import io
+import re
 import sys
+import struct
 from pathlib import Path
 from importlib import import_module
-from importlib.abc import Loader, MetaPathFinder
+from importlib.abc import Loader, MetaPathFinder, ResourceReader
 from importlib.machinery import ModuleSpec
 
 from wasmtime import Module, Linker, Store, WasiConfig
 from wasmtime import Func, Table, Global, Memory
+from wasmtime import wat2wasm, bindgen
 
 
 predefined_modules = []
@@ -28,7 +33,10 @@ linker.define_wasi()
 linker.allow_shadowing = True
 
 
-class _WasmtimeLoader(Loader):
+_component_bindings: Dict[Path, Mapping[str, bytes]] = {}
+
+
+class _CoreWasmLoader(Loader):
     def create_module(self, spec):  # type: ignore
         return None  # use default module creation semantics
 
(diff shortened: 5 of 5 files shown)


whitequark:
This PR now includes an example of using autogenerated bindings for components.

whitequark:
That's a lot of mypy failures. I'm unconvinced mypy is actually useful given the sheer amount of `Any` and `# typing: ignore` annotations all this requiers but oh well...

whitequark:
That was a hour of my life I'll never get back but at least mypy is happy now. (Having to spend this much effort on busywork actually kind of turns me off contributing nontrivial code to wasmtime-py so I think next time I'll just go with `# type: ignore` until tests pass; every single change that I've made in the last hour was a false positive of some sort, in that it did not point to an error in my code.)

Aside from that, I think this is ready to merge now, with two caveats:
- This is still not thread-safe. I could easily make it so, but it's not totally clear to me what the granularity should be.
   - Store per thread? Works for core modules but it gets really weird when two unrelated libraries both use `wasmtime.loader` and their stuff ends up in the same store now. If one of them hits an unrecoverable trap I think the other will stop working too, maybe?
   - Store per `.wasm` file? Works for components but I *think* that wouldn't really work with the existing core module import code?
   - Store per `.wasm` file, except if you import something you use the import's store? Works for the case where your entire dependency tree eventually depends on the same module, which might actually not be uncommon, but doesn't work in general.
   - Some way to indicate which store to use?
   - Store per `.wasm` file, and all imports are imported anew? Works universally but will break mutable aliasing.
- The interface differs between importing core modules and components, with the latter requiring an explicit `store` argument.

Ultimately I feel like maybe we should break the interface and surface the `store` argument explicitly for all use cases, core and component, because it's not obvious to me that a general solution is possible, or even a solution general enough to make most downstream users happy. And there is clearly unhappiness related to the status quo, such as https://github.com/bytecodealliance/wasmtime-py/issues/223.

whitequark:
Rebased.

whitequark:
Note that I've added an example but I'm not totally sure if the .wat file I committed follows the best practices for that kind of thing. I did a `wasm-tools print` and then hand-edited it to be more legible and closer to existing .wat files but a lot of the Component Model internals are still somewhat opaque to me.

In any case I hope the example provides a reasonable amount of testing for this functionality and also shows how to use it.

I basically don't show how to handle imports at all (which is unlike the core module handling code) and this is mainly because I don't expect to need more than WASI imports in close future at least, possibly ever; I think maybe another contributor is better positioned to address that, though if it's not a high-effort change I could do something as well.

rust/src/bindgen.rs
```diff
@@ -538,16 +538,16 @@ impl<'a> Instantiator<'a> {
...
         uwriteln!(
             self.gen.init,
-            "path = pathlib.Path(__file__).parent / ('{}')",
+            "file = importlib_resources.files() / ('{}')",
             core_file_name,
         );
         uwriteln!(
             self.gen.init,
-            "module = wasmtime.Module.from_file(store.engine, path)"
+            "module = wasmtime.Module(store.engine, file.read_bytes())"
```
alexcrichton, line 550: Is it possible to test whether `file` exists on the filesystem? Being able to plumb the file-ness all the way down can sometimes be useful in terms of various optimizations. In this case I don't think it changes anything today, but it might enable Wasmtime to do various tricks in the future.

alexcrichton (approved):
Looks great to me, thank you for this!

> Having to spend this much effort on busywork actually kind of turns me off contributing nontrivial code to wasmtime-py

Sorry about that! I originally thought this would be a good place to have type annotations but I'm not sure if I'm swimming upstream here. I don't want to make this library hard to use or contribute to at all, but I figured that if folks had their own type annotations they'd want the usage here checked as well. I do agree though that the more "dynamic" parts like the loader I think can probably skip type-checks entirely.

> This is still not thread-safe

I agree this is pretty thorny, and I agree it's best to defer to later. I think your idea is the right way to go here of surface a different API for core-wasm-modules and always require a `Store` is passed in. There's basically no way to share an instance across threads, and if a normal Python module can be shared across threads then representing a wasm instance as a Python module isn't the right way to go.

> I'm not totally sure if the .wat file I committed follows the best practices for that kind of thing

Looks good to me!

> I basically don't show how to handle imports at all

I think deferring this to later is fine yeah. The loader-based pieces were more demo-quality when I first added them anyway so I think it's fine to figure out how to best manage this later.

alexcrichton:
Do you have a sense for if there's a more conventional type-checker to use? I picked mypy as it seemed like a reasonable default at the time, but I could very well have been wrong and/or times may have changed in the meantime. I've got no love of mypy myself as I've run into lots of issues historically related to `ctypes` and mypy, so if Pyright/Pylance are better options I'd be happy to switch to that.

I definitely agree that mypy has felt like a lot of busywork, especially in all the `-> None` functions that don't return anything. I'd just assumed it was The Way up until now!

whitequark:
So, full disclosure, I'm mostly using Pylance from within VS Code, so my user experience will not quite match that of contributors using command-line Pyright analysis. However, I just ran Pyright on the `wasmtime/loader.py` file that I've been editing, here's the output:

```
$ ./node_modules/.bin/pyright -v .venv/ wasmtime/loader.py 
.../wasmtime/loader.py
  .../wasmtime/loader.py:44:54 - error: Argument of type "str | None" cannot be assigned to parameter "path" of type "str | bytes | PathLike[Unknown]" in function "from_file" (reportArgumentType)
  .../wasmtime/loader.py:44:70 - error: "origin" is not a known member of "None" (reportOptionalMemberAccess)
  .../wasmtime/loader.py:52:20 - error: Argument of type "str | None" cannot be assigned to parameter "key" of type "str" in function "__getitem__"
    Type "str | None" cannot be assigned to type "str"
      "None" is incompatible with "str" (reportArgumentType)
  .../wasmtime/loader.py:54:36 - error: Argument of type "AsExternType" cannot be assigned to parameter "ty" of type "FuncType" in function "__init__"
    Type "AsExternType" cannot be assigned to type "FuncType"
      "GlobalType" is incompatible with "FuncType" (reportArgumentType)
  .../wasmtime/loader.py:55:47 - error: Argument of type "str | None" cannot be assigned to parameter "name" of type "str" in function "define"
    Type "str | None" cannot be assigned to type "str"
      "None" is incompatible with "str" (reportArgumentType)
  .../wasmtime/loader.py:74:23 - error: Argument of type "str | None" cannot be assigned to parameter "args" of type "StrPath" in function "__new__" (reportArgumentType)
  .../wasmtime/loader.py:74:39 - error: "origin" is not a known member of "None" (reportOptionalMemberAccess)
  .../wasmtime/loader.py:94:9 - error: Method "is_resource" overrides class "ResourceReader" in an incompatible manner
    Parameter 2 name mismatch: base parameter is named "path", override parameter is named "name" (reportIncompatibleMethodOverride)
8 errors, 0 warnings, 0 informations 
```

Not only are these messages much more readable, but some of them are actionable, too! Let's do the breakdown of these 8 messages:

1. `error: Argument of type "str | None" cannot be assigned to parameter "path" of type "str | bytes | PathLike[Unknown]" in function "from_file"`: false positive, this is dynamically impossible (but it doesn't know this and this could have easily been a bug)
2. `error: "origin" is not a known member of "None"`: continuation of the same
3. `error: Argument of type "str | None" cannot be assigned to parameter "key" of type "str" in function "__getitem__"`: seems like a bug: unnamed import types would crash the loader
4. `error: Argument of type "AsExternType" cannot be assigned to parameter "ty" of type "FuncType" in function "__init__"`: seems like a bug: insufficiently generic type annotation
5. `error: Argument of type "str | None" cannot be assigned to parameter "name" of type "str" in function "define"`: seems like a bug: unnamed export types would crash the loader
6. `error: Argument of type "str | None" cannot be assigned to parameter "args" of type "StrPath" in function "__new__"`: same as (1)
7. `error: "origin" is not a known member of "None"`: continuation of (6)
8. `error: Method "is_resource" overrides class "ResourceReader" in an incompatible manner
    Parameter 2 name mismatch: base parameter is named "path", override parameter is named "name"`: definitely a bug, if someone uses a kwarg to call this method (unlikely but possible), it'll crash

So, out of the 6 errors reported as 8 diagnostics, only 2 are false positives due to runtime invariants outside of the typechecker's ability to reason, 2 are definitely bugs that could be reasonably triggered through normal use, and 2 are bugs that could only be triggered through the use of the module linking proposal extensions. There aren't any false positives due to insufficiently powerful analyses or the like. Mypy, of course, found none of the 6.

To handle the false positives due to runtime invariants all I had to do was to add

    assert module.__spec__ is not None and module.__spec__.origin is not None

and now they're gone. The check doesn't hurt readability either.

whitequark:
@alexcrichton OK, it should be calling `from_file` now for the case where it's actually on the filesystem.

alexcrichton:
Nice! I'll try to look into switching to pylance/pyright in the future

PR #228, opened 2024-04-15
Update to latest bindings

alexcrichton:
Handle some updates related to externref and change how `Val` works by making it no longer store `wasmtime_val_t` but instead store the raw value itself which is converted to `wasmtime_val_t` on-the-fly.


alexcrichton:
Test failures here are https://github.com/bytecodealliance/wasmtime/issues/8367

Issue #202, opened 2024-01-11
wasmtime bindgen `get_terminal_stdin` NotImplementedError when using WIT component


ConradMearns:
I'm attempting to follow along with https://github.com/bytecodealliance/componentize-py to write a python WASM component that returns "Hello World!", and import that WASM module into python to execute it.

I have a `hello.wit`
```wit
package example:hello;

world hello {
  export hello: func() -> string;
}
```

`hello.py`
```python
import hello
class Hello(hello.Hello):
    def hello(self) -> str:
        return "Hello world!"
```

I can build the WASM file with
```bash
componentize-py -d hello.wit hello bindings .
componentize-py -d hello.wit -w hello componentize app -o hello.wasm
```

but when I try to start the process of utilizing this module, I run into an error with `wasmtime.bindgen`

```bash
poetry run python -m wasmtime.bindgen hello.wasm --out-dir hellowasm
```

```
Traceback (most recent call last):
  File "/home/conrad/test_wasmwit/.venv/lib/python3.11/site-packages/wasmtime/_func.py", line 260, in enter_wasm
    yield byref(trap)
  File "/home/conrad/test_wasmwit/.venv/lib/python3.11/site-packages/wasmtime/_func.py", line 101, in __call__
    raise WasmtimeError._from_ptr(error)
wasmtime._error.WasmtimeError: error while executing at wasm backtrace:
    0:  0x100 - wit-component:shim!indirect-wasi:cli/terminal-stdin-get-terminal-stdin
    1: 0x254f - wit-component:adapter:wasi_snapshot_preview1!_ZN22wasi_snapshot_preview111descriptors11Descriptors3new17ha2fdde51d30a71bdE
    2: 0x1640 - wit-component:adapter:wasi_snapshot_preview1!_ZN22wasi_snapshot_preview15State11descriptors17h780c546b61bcfc7fE
    3: 0x180c - wit-component:adapter:wasi_snapshot_preview1!fd_write
    4:  0x13a - wit-component:shim!adapt-wasi_snapshot_preview1-fd_write
    5: 0x24c210 - <unknown>!_ZN4wasi13lib_generated8fd_write17hd4964fea612b930fE
    6: 0x2487fd - <unknown>!_ZN80_$LT$std..io..Write..write_fmt..Adapter$LT$T$GT$$u20$as$u20$core..fmt..Write$GT$9write_str17h375f1d6863bea9dfE
[...]

Issue #197, opened 2023-11-10
Support WIT resource types


landonxjames:
I appears that having a wit `resource` type in a component definition causes wasmtime-py to throw an error when generating bindings. 

The following .wit file:
```wit
package project:runtime

interface clients {
  resource placeholder {
    hello-world: func() -> string
  }
}

world runtime {
  export clients
}
```
Implemented by this Rust code:
```rust
//WIT imports
wit_bindgen::generate!({
world: "runtime",
exports: {
    "project:runtime/clients/placeholder": Placeholder,
}});
use exports::project::runtime::clients::GuestPlaceholder;

pub struct Placeholder;

impl GuestPlaceholder for Placeholder {
    fn hello_world(&self) -> String {
        let val = "Hello World";
        val.to_string()
    }
}
```

and compiled using:
* Cargo - 1.73.0
* Wasm-tools - 1.0.51
* Wasmtime-py - 14.0.0

Throws the following error:

<details><summary>Full Error</summary>
<p>

```sh
$ wasm-tools component new ./target/wasm32-unknown-unknown/release/wasm_runtime.wasm -o runtime_component_resources.wasm

$ python -m wasmtime.bindgen runtime_component_resources.wasm --out-dir pybind/
Traceback (most recent call last):
  File "/Volumes/workplace/project/.venv/lib/python3.11/site-packages/wasmtime/_func.py", line 260, in enter_wasm
    yield byref(trap)
  File "/Volumes/workplace/project/.venv/lib/python3.11/site-packages/wasmtime/_func.py", line 101, in __call__
    raise WasmtimeError._from_ptr(error)
wasmtime._error.WasmtimeError: error while executing at wasm backtrace:
    0:  0x100 - wit-component:shim!indirect-wasi:cli/terminal-stdin-get-terminal-stdin
    1: 0x254f - wit-component:adapter:wasi_snapshot_preview1!wasi_snapshot_preview1::descriptors::Descriptors::new::ha2fdde51d30a71bd
    2: 0x1640 - wit-component:adapter:wasi_snapshot_preview1!wasi_snapshot_preview1::State::descriptors::h780c546b61bcfc7f
    3: 0x180c - wit-component:adapter:wasi_snapshot_preview1!fd_write
    4:  0x13a - wit-component:shim!adapt-wasi_snapshot_preview1-fd_write
    5: 0x1b080d - <unknown>!wasi::lib_generated::fd_write::hd4964fea612b930f
[...]

PR #232, opened 2024-05-07
Initial implemention of bindgen for exported resources

jamesls:
Issue #197 

This adds support for generating bindings for resources exported from a
wasm component, so that Python code can interact with these resources.

I looked over jco/componentize-py and followed the same overall
approach.  The internal structuring of `bindgen.rs` here is slightly
different than jco, so I erred on the side of keeping things as they are
as much as possible.

In terms of the generated code itself:

* A resource maps to a Python class.
* Calling the constructor of a resource returns a new instance of a
  Python class.
* Protocols are generated for each resource in the module corresponding
  to the exported interface.
* The protocols follow the existing calling convention of
  `func(caller: Store, ...<modeled args>) for exported functions.
* The instantiation of the resource class is accessible from the instance
  of the class that holds the scope for the exported interface.

One of the issues I had to work out was how to pass in the root object
that contains all the trampolines needed by the resource classes, which
is defined in the root object.  This is an issue because the
`constructor` of a resource is modeled in WIT, and the root object is
typically passed as a param in the `__init__`.  What I ended up doing
was using a closure to capture the root component instance and then
dynamically generating the class when the interface "scope class" is
instantiated.  For example, given this WIT file:

```wit
package component:basicresource;

interface my-interface-name {
    resource demo-resource-class {
        constructor(name: string);
        greet: func(greeting: string) -> string;
    }
}

world example {
    export my-interface-name;
}
```

The interface module for `my-interface-name` would have:

```python

class MyInterfaceName:
    component: 'Root'
    
    def __init__(self, component: 'Root') -> None:
        self.component = component
        self.DemoResourceClass = _create_demo_resource_class(component)

# ...

def _create_demo_resource_class(component: 'Root') -> type[DemoResourceClass]:
    class _DemoResourceClass:
        # Preserve the signature defined in WIT.
        def __init__(self, caller: wasmtime.Store, name: str) -> None:
            # Capture the root component for access to all the trampolines/canons.
            self.component = component
```

And the code to instantiate a resource looks like this:

```python
from wasmtime import Store

import mybindings # Whatever name you used.


store = Store()
root = mybindings.Root(store)
interface = root.my_interface_name()
instance = interface.DemoResourceClass(store, "myname")
print(instance.greet(store, "Hello"))
```

Would love feedback on this, it's unconventional in Python, but I suppose it
also maps closer to the idea of a`(resource ...)` being a type constructor
that creates a fresh type for each instance, while still being able to use the
the module level protocols for the resource to satisfy type checking of the
resources.

Imported resources aren't implemented as part of this change.  Trying to
bindgen imported resources will still result in an `unimplemented!()`
panic.  Wanted to get feedback on the overall approach before going
too far.

As part of this change I also added a commit to implement more of the WASI
shims needed to run bindgen (same stack trace from #202).


whitequark (approved):
The class-in-function idiom looks fine to me. It's not *common* but I wouldn't say it's *unidiomatic* either; just rarely used.

One remark on it is that the name of the function ends up a part of the qualified name of the class, so you might want to name that something sensible because it will end up in error messages and debug output everywhere.

rust/src/bindgen.rs
```diff
@@ -531,8 +573,109 @@ impl<'a> Instantiator<'a> {
...
+                    uwriteln!(
+                        self.gen.init,
+                        "_resource_drop_{resource_id}_ty = wasmtime.FuncType([wasmtime.ValType.i32()], [])"
+                    );
+                    uwriteln!(
+                        self.gen.init,
+                        "trampoline{tidx} = wasmtime.Func(store, _resource_drop_{resource_id}_ty, _resource_drop_{resource_id})"
+                    );
+                }
+                Trampoline::ResourceRep(_) => {}
```
alexcrichton, line 616: Could this be `unreachable!` if it's left unimplemented?

alexcrichton (approved):
Thanks so much for helping to work on this! I've only got one minor comment below, but otherwise I think this all looks quite reasonable to me and would be happy to merge.

The main thing I might recommend perhaps bikeshedding a bit on is the testing story here. Tests so far in this repository are raw inline text-format-components which aren't exactly readable or easy to work with. Do you have thoughts/ideas on how to improve this? For example it might be reasonable to use componentize-py here to help generate test cases?

Ideally I'd imagine that test cases are a WIT world plus a "thing" which is a component, where the "thing" is probably ideally Python given this repository but could be anything that compiles to a component. That would make things much more readable and easier to write tests I think

jamesls:
> The main thing I might recommend perhaps bikeshedding a bit on is the testing story here. Tests so far in this repository are raw inline text-format-components which aren't exactly readable or easy to work with. Do you have thoughts/ideas on how to improve this? For example it might be reasonable to use componentize-py here to help generate test cases?

Interesting idea, I'll try and sketch out writing tests using componentize-py and see how it goes.  If it works out, it would be easier to write tests compared to writing wat.

alexcrichton:
I'm really liking how #234 is shaping up, thanks again for that! In the meantime no need to block this on that, so I'm going to go ahead and merge this. Thank you again for your work here!

PR #255, opened 2024-10-23
Wiring SharedMemory with it's corresponding bindings

atilag:
I think this is more or less what we need to have an initial support for SharedMemory.
I'll be adding tests and examples if you think I'm heading into the right direction.

TODO:
- [x] Add tests for Shared Memory
- [x] Adapt the rest of the bindings to the new c-api changes


wasmtime/_ffi.py
```diff
@@ -53,6 +53,7 @@
 WASMTIME_EXTERN_GLOBAL = c_uint8(1)
 WASMTIME_EXTERN_TABLE = c_uint8(2)
 WASMTIME_EXTERN_MEMORY = c_uint8(3)
+WASMTIME_EXTERN_SHAREDMEMORY = c_uint8(4)
```
atilag, line 56: This was wierd, because the `4` is shared with the definition below (`WASMTIME_EXTERN_INSTANCE = c_uint8(4)`) but `4` is also how wasmtime runtime defines this const.

wasmtime/_sharedmemory.py
```diff
@@ -0,0 +1,149 @@
...
+        np_mem = np.frombuffer(memory.get_buffer_ptr(store), dtype=np.uint8)
+        np_mem[start:end] = A # write
+        B = np_mem[start:end] # read
+        """
+        if size is None:
+            size = self.data_len(store)
+        ptr_type = ctypes.c_ubyte * size
+        return ptr_type.from_address(ctypes.addressof(self.data_ptr(store).contents) + offset)
+
+    def read(
```
atilag, line 84: This has been copied straight from the Memory code and I didn't put too much attention just yet...

wasmtime/_sharedmemory.py
```diff
@@ -0,0 +1,149 @@
...
+        key = slice(start, stop, None)
+        start, stop, _ = key.indices(size)
+        val_size = stop - start
+        if val_size <= 0:
+            # return bytearray of size zero
+            return bytearray(0)
+        src_ptr = self.get_buffer_ptr(store, val_size, start)
+        return bytearray(src_ptr)
+
+    def write(
```
atilag, line 106: same as `read` -- copied from the Memory code and not reviewed it thoroughly.

atilag:
I have a "test case" with this code:
```python
        import wasmtime

        engine = wasmtime.Engine()

        wasi_config = wasmtime.WasiConfig()
        wasi_config.inherit_stdout()

        store = wasmtime.Store(engine)
        store.set_wasi(wasi_config)

        wasm_file = 'my_wasi_program.wasm'
        with open(wasm_file, 'rb') as f:
            wasm_bytes = f.read()

        module = wasmtime.Module(engine, wasm_bytes)

        linker = wasmtime.Linker(engine)
        linker.define_wasi()

        memory_type = wasmtime.MemoryType(wasmtime.Limits(0, 4000), shared=True)
        memory = wasmtime.SharedMemory(engine, memory_type)

      
        linker.define(store, "env", "memory", memory) 


        instance = linker.instantiate(store, module)
        ...
```
This is throwing an error coming from the wasmtime runtime:
```
ty = <wasmtime._types.MemoryType object at 0x101e30580>

    def __init__(self, engine: Engine, ty: MemoryType):
        """
        Creates a new shared memory in `store` with the given `ty`
        """

        sharedmemory = ffi.wasmtime_sharedmemory_t()
        ptr = POINTER(ffi.wasmtime_sharedmemory_t)()
        error = ffi.wasmtime_sharedmemory_new(engine.ptr(), ty.ptr(), byref(ptr))
        if error:
>           raise WasmtimeError._from_ptr(error)
E           wasmtime._error.WasmtimeError: shared memory must have the `shared` flag enabled on its memory type
```
which kind of make sense, because there's no support for the "shared" flag in the `wasmtime_memorytype_new` exported in the C-API. I also have a patch a there (will PR soon), and made some progress, but then a clone() of the shared memory type is panicking, so still trying to figure out why.

wasmtime/_bindings.py
```diff
@@ -2937,9 +2937,9 @@ def wasmtime_linker_instantiate_pre(linker: Any, module: Any, instance_pre: Any)
 
 _wasmtime_memorytype_new = dll.wasmtime_memorytype_new
 _wasmtime_memorytype_new.restype = POINTER(wasm_memorytype_t)
-_wasmtime_memorytype_new.argtypes = [c_uint64, c_bool, c_uint64, c_bool]
-def wasmtime_memorytype_new(min: Any, max_present: Any, max: Any, is_64: Any) -> ctypes._Pointer:
-    return _wasmtime_memorytype_new(min, max_present, max, is_64)  # type: ignore
+_wasmtime_memorytype_new.argtypes = [c_uint64, c_bool, c_uint64, c_bool, c_bool]
+def wasmtime_memorytype_new(min: Any, max_present: Any, max: Any, is_64: Any, shared: Any) -> ctypes._Pointer:
+    return _wasmtime_memorytype_new(min, max_present, max, is_64, shared)  # type: ignore
```
alexcrichton, line 2942: FWIW this `bindings.py` is auto-generated meaning that you can't modify it directly. This reflects the headers themselves and I think it's a bug in the C API that you can't create a new shared memory type. I think that a function such as `wasmtime_sharedmemorytype_new` will need to be added to the C API (or a `shared: bool` flag to the existing constructor here)

wasmtime/_ffi.py
alexcrichton: Good catch! The `WASMTIME_EXTERN_{INSTANCE,MODULE}` types should actually be removed since the C API hasn't had support for those in quite awhile.

wasmtime/_sharedmemory.py
```diff
@@ -0,0 +1,149 @@
...
+from ctypes import *
+import ctypes
+import typing
+from wasmtime import MemoryType, WasmtimeError, Engine
+from ._store import Storelike
+
+
+
+class SharedMemory:
+    _sharedmemory: ffi.wasmtime_sharedmemory_t
```
alexcrichton, line 11: Could this have a superclass of `Managed` like most other types in the bindings?

wasmtime/_sharedmemory.py
alexcrichton: I think it looks reasonable enough but you'll be able to remove the `store` argument

alexcrichton: It may also be worthwhile extracting this to a helper to be shared between shared and non-shared memory?

wasmtime/_bindings.py
atilag: Yes, I had to add the `bool shared` there and some more glue code that was missing. Let me PR there and restore this file.

wasmtime/_sharedmemory.py
atilag: I just realized that the read/write methods and their respective supporting methods, rely on the Store, but the SharedMemory doesn't seem to need a Store. So I wento the Rust API implementation in wasmtime and saw that there are no read/write methods in there, which I guess is aligned with the security model (or a lack of standarization in the IO for shared memory just yet).
I have removed the read/write methdos and their respective supporing methdos.

wasmtime/_sharedmemory.py
atilag: Done!
Let me know if I did something wrong

wasmtime/_sharedmemory.py
```diff
@@ -0,0 +1,74 @@
...
+        ptr = POINTER(ffi.wasmtime_sharedmemory_t)()
+        error = ffi.wasmtime_sharedmemory_new(engine.ptr(), ty.ptr(), byref(ptr))
+        if error:
+            raise WasmtimeError._from_ptr(error)
+        self._set_ptr(sharedmemory)
+
+    @classmethod
+    def _from_raw(cls, sharedmemory: ffi.wasmtime_sharedmemory_t) -> "SharedMemory":
+        ty: "SharedMemory" = cls.__new__(cls)
+        ty._set_ptr(sharedmemory)
```
atilag, line 25: This is correct, right?

alexcrichton: Looks good! I might recommend also adding a runtime type-check [like with traps](https://github.com/bytecodealliance/wasmtime-py/blob/157e307aec8e864159181312ecee4e8c283b3c51/wasmtime/_trap.py#L49-L50) to help protect against unsafety at runtime

alexcrichton:
Looks good to me! I think the lack of read/write on the Rust side of things is probably an oversight and so might be worth adding here eventually, but I think it's nonetheless reasonable to start here. Before landing this'll want to have a few smoke tests too to ensure that everything is wired up right

wasmtime/_sharedmemory.py
atilag: hmmm... you mean something like:
```python
if not isinstance(sharedmemory, ffi.wasmtime_sharedmemory_t):
   raise TypeError(...)
```
?

atilag:
Do you know  how can I generate the `_bindings.py` file out of the `wasmtime` C-API? I have the latter cloned, modified to add the 5th argument (bool shared) to the MemoryType, and I'm able to build an generate the .dlyb (on my mac).

alexcrichton:
It's unfortunately not easy at this time. Basically `./ci/download-wasmtime.py` will place all the headers/etc into a location and what you'll want to do is overwrite/link the in-tree headers in the same location. From there you can also replace the `*.dylib` with what you built locally, and that way you can test locally.

atilag:
Ok, I'm resuming this, now that we have all the pieces in place for the wasmtime runtime.

alexcrichton:
Sounds good! As a heads up CI here pulls from a release branch of Wasmtime but you should be able to pull it from the "dev" release which should have all the latest changes.

atilag:
After pulling lastest changes from the c-api and regenerating the bindings, I realized that I need to integrate other changes not related to the memory as well (because regenerating the bindings pull all the changes, of course). This is fine, I can do it here as well, I wouldn't expect the PR to grow a lot.

alexcrichton:
Ah yeah no worries, happy to review that here too!

alexcrichton:
Feel free to hack around things here and/or split out the update to Wasmtime 27. I'm busy this week but I can more thoroughly dig in next week

atilag:
I'm having a Seg. Fault after writing a test, still debugging it (been busy this past weeks!). Once I have fixed it, I'll update the branch and I think we should be ok to merge this.

alexcrichton:
It looks like mypy is reporting some type errors which might be the reason for the faults? If mypy passes it may be easier to debug perhaps

atilag:
Yep, mypy has indeed captured all the problems.
I added some unit tests too. I used the same testing structure that exists around Memory, it's a good start but maybe looks insuficient. Will try to add some acceptance tests in another follow up PR, with the intention to merge this asap if you find it ready.
Thanks Alex!

alexcrichton:
Looks good to me! Could you go ahead and add the tests here? I'd like to make sure to at least exercise some basics in CI even if more advanced testing is coming later

atilag:
Of course that I forgot to git add the test file xDDD
Done!

alexcrichton (approved):
Thanks again for this! Would you like a publish of 27.0.2 for this?

atilag:
I don't have a "production" use-case for this just yet, we are investigating the potential fit for some use-cases we have, but releasing would make things a little bit easier moving forward, so it's up to you :)

alexcrichton:
Sounds good! I'll do that in https://github.com/bytecodealliance/wasmtime-py/pull/261

PR #257, opened 2024-11-15
Update to latest wasmtime dev

### LINKED ISSUE #251: Add `dir_perms` and `file_perms` to `WasiConfig.preopen_dir`
The rust [preopened_dir][1] method has parameters for setting permissions to allow mapping the directory or its files as read-only. The python `preopen_dir` doesn't currently expose these parameters, so we can only map directories as read-write.

Is it possible to add the ability to set permissions on mapped directories?

[1]: https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/struct.WasiCtxBuilder.html#method.preopened_dir

jder:
Closes #251 

As discussed in there:
* Updates to latest dev wasmtime and latest wasmtime release (26) for rust bindgen code
* Translates enums in c headers into python-accessible values.
* Wraps new preopen file/directory permissions into a public python API.
* A few other minor changes to accommodate dev changes (32 -> 64 bits, a new "shared" flag)

There is more we could do using the exposed enums (in existing bindings where we have numbers now) and exposing new APIs but thought we'd land this first.

### DIFF
--- a/ci/cbindgen.py
+++ b/ci/cbindgen.py
@@ -20,6 +20,7 @@ class Visitor(c_ast.NodeVisitor):
         self.ret += 'from ctypes import *\n'
         self.ret += 'import ctypes\n'
         self.ret += 'from typing import Any\n'
+        self.ret += 'from enum import Enum, auto\n'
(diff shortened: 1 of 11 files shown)


alexcrichton:
Thanks!

PR #265, opened 2025-01-16
WIP: add wasi-threads support

thewtex:


thewtex:
Per discussion in https://github.com/bytecodealliance/wasmtime/pull/10030 , the implementation for this use case is likely not `wasi-threads`. But, the test here is relevant to real-world use and adoption of wasm/wasi/wasmtime, so I am leaving this open.

Issue #247, opened 2024-07-16
Missing Windows ARM64 support


ben-zen:
I was setting up my new Glasgow, and as I went to build the demo, I discovered that this package doesn't have a Windows AARCH64 build.

Looking deeper, I found a few [upstream](https://github.com/bytecodealliance/wasmtime/issues/4992) [issues](https://github.com/bytecodealliance/wasmtime/issues/8247) that are blocking forward motion. Unfortunately, I'm not familiar with unwinding in Rust to be able to help. Once wasmtime has Windows on ARM64 support, it'd be swell to get wasmtime-py support for the platform.

alexcrichton:
Thanks for the report! Support upstream is indeed required first, so this'll be blocked on the issues you linked.

billti:
I see there is a binary neutral wheel at https://pypi.org/project/wasmtime/#files . This is what I got when I tried to 'pip install wasmtime' in Python on Windows ARM64. I thought maybe it would do a source install, however it just seems to install the x64 library and fails with "unsupported architecture" when imported. (See screen shots below)

<img width="697" alt="Image" src="https://github.com/user-attachments/assets/2a4c0fb6-329a-4d0a-ae46-b1e95d3742fa" />

<img width="643" alt="Image" src="https://github.com/user-attachments/assets/7606705a-a4ae-4940-9d06-27dc4510268c" />

I see the above 'upstream issue' mentioned was resovled a while back. Cloning the repo and doing a 'cargo build --release' works fine. Should this be resolved now? Any specific validation would need to be done? Any way I can help get Windows on ARM64 binaries/wheels supported? Thanks.

cfallin:
Reading the [code](https://github.com/bytecodealliance/wasmtime-py/blob/0b66b4f50fa6d45f9a981d9c6c2537e81f14c5a0/wasmtime/_ffi.py#L21-L26) that throws the error, I see

```
if machine == 'AMD64':
    machine = 'x86_64'
if machine == 'arm64':
    machine = 'aarch64'
if machine != 'x86_64' and machine != 'aarch64':
    raise RuntimeError("unsupported architecture for wasmtime: {}".format(machine))
```

so we accept `arm64` (macOS gives this) and `aarch64` (Linux) but not `ARM64` as seen in the exception's error message. It may be as simple as adding a case here -- Alex above is the main maintainer but I imagine PRs are welcome for this?

billti:
[...]

PR #268, opened 2025-02-07
Support Windows on ARM64

billti:
This is for issue #247 . Putting up for early review, but some testing on other platforms and a couple questions remain.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,24 +79,14 @@ jobs:
         python setup.py bdist_wheel --plat-name win-amd64
     - run: |
         git clean -fdx wasmtime build
+        python ci/download-wasmtime.py win32 arm64
+        python ci/build-rust.py
+        python setup.py bdist_wheel --plat-name win-arm64
+    - run: |
+        git clean -fdx wasmtime build
         python ci/download-wasmtime.py musl x86_64
         python ci/build-rust.py
         python setup.py bdist_wheel --plat-name musllinux_1_2_x86_64
-
-    # Build an "any" wheel with:
-    #
-    #   * MinGW
-    #
-    # because at this time I don't know what the `--plat-name` tags supported on
-    # PyPI are for these platforms. Our hope is that any platform not matching
-    # the above `--plat-name` arguments will install this `any` wheel instead,
-    # and then when the wheel runs it'll dynamically select from the available
-    # shared libraries.
-    - run: |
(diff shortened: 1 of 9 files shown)


billti:
(Looks like the CI has been failing for a couple days unrelated to my change)

billti:
Actually I see in the runner image used for Windows (https://github.com/actions/runner-images/blob/win22/20250127.1/images/windows/Windows2022-Readme.md) that GCC is included. So I'll revert that change to using 'clang' on Windows, install that locally, and try again.

billti:
With the latest couple of commits and with 'gcc' installed locally via MSYS2, all tests are passing and the wheel is building successfully!

alexcrichton (approved):
Thank you for this!

alexcrichton:
CI failures here will get fixed in https://github.com/bytecodealliance/actions/pull/15 (or so I'm hoping), so I'll take care of the CI bits for this in the near future.

alexcrichton:
Ok I think that CI failure might be legitimate?

billti:
Ugh. As well as platform.machine() not matching on arm64, looks like Windows has a different value to Linux/Mac for x64 as well (on Mac and Linux it's `x86_64`). Sigh...

```
C:\code>python -c "import platform;print(platform.system());print(platform.machine())"
Windows
AMD64
```

I've just realized that this is also going to hit the same problem we have in our `qsharp` project as well, in that until there are Windows ARM64 runners available in the CI/release pipelines, while you can cross-compiler for Windows ARM64 fine, you won't be able to run the tests. We decided to live without test coverage for Windows ARM64 in our pipelines (https://github.com/microsoft/qsharp/blob/main/.ado/publish.yml#L343) and spot check the releases manually until Windows ARM64 runners become available, but I'll leave it up to you if that's acceptable here.

I'll push a commit shortly that fixes x64 on Windows and removes Windows aarch64 from the pipeline until you comment on the test coverage question. Thanks!

billti:
I noticed none of the tests run on arm64 for any platforms - so there was nothing to remove and it seems in line. CI is green now! As extra validation, I copied/installed the wheel for the CI build to my Windows ARM64 machine into a clean venv and ran a few of the scripts from under ./examples and they all run fine.

alexcrichton:
Thanks for confirming!

And yeah we could probably set up and arm64 runner now that ubuntu-arm64 images exist on github actions, but alas I don't think there are windows arm64 images just yet...

PR #285, opened 2025-06-10
Build source tar.gz (sdist) alongside wheels (bdist_wheel)

moreati:
This is a first pass. Probably enough to build conda packages, but not enough for pip or uv to build from. I've tried to be minimally invasive

- Keep setuptools as the packaging backend
- Keep `setup.py` as the source of truth

A more complete overhaul might consider moving to a more recent/focused build library (e.g. PyO3, Maturin); and/or a more recent (e.g. `setup.cfg`, `pyproject.toml`).

A diff of file listings before and after show the wheel contents are unchanged, except for version specific paths

```console
$ diff -u \
  <(zipinfo -1 dist-before/wasmtime-33.0.0.dev362-py3-none-macosx_11_0_arm64.whl|sort) \
  <(zipinfo -1 dist.after/wasmtime-33.0.0.dev363-py3-none-macosx_11_0_arm64.whl|sort)
--- /dev/fd/11  2025-06-10 09:53:26
+++ /dev/fd/12  2025-06-10 09:53:26
@@ -1,8 +1,8 @@
-wasmtime-33.0.0.dev362.dist-info/LICENSE
-wasmtime-33.0.0.dev362.dist-info/METADATA
-wasmtime-33.0.0.dev362.dist-info/RECORD
-wasmtime-33.0.0.dev362.dist-info/top_level.txt
-wasmtime-33.0.0.dev362.dist-info/WHEEL
+wasmtime-33.0.0.dev363.dist-info/LICENSE
+wasmtime-33.0.0.dev363.dist-info/METADATA
+wasmtime-33.0.0.dev363.dist-info/RECORD
+wasmtime-33.0.0.dev363.dist-info/top_level.txt
+wasmtime-33.0.0.dev363.dist-info/WHEEL
 wasmtime/__init__.py
 wasmtime/_bindings.py
 wasmtime/_config.py
```

refs #231

PR #287, opened 2025-06-13
Use Python "build" package as build frontend, don't directly run setup.py

### LINKED ISSUE #231: add source distribution to pypi - to create conda package
Pypi distribution is missing source files - 
https://pypi.org/project/wasmtime/#files
Adding the source will help create a conda package, and make wasmtime available for conda users.

### LINKED ISSUE #286: CI jobs call setup.py directly, this will be unsupported from 2025-10-31
.github/workflows/main.yml run setup.py as a command in several places. This mode of operation has been deprecated by the setuptools project for some time,. It will not be supported after 31 Oct 2025. This is much sooner than I'd realised.

AIUI having a setup.py and using setuptools as a build backend will still be supported, through the newer [build](https://github.com/pypa/build).

Warning seen

```
/Users/alex/src/wasmtime-py/.venv/lib/python3.13/site-packages/setuptools/_distutils/cmd.py:90: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        By 2025-Oct-31, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

!!
  self.initialize_options()
```

refs #231, #285

moreati:
Running setup.py directly is deprecated, from 2025-10-31 it won't be supported. The Python ecosystem is moving toward seperation of build frontend and backend. This commit switches to the "build" frontend, and keeps setuptools as the backend.

~~Two~~ three `setup()` keywords have been altered

`packages` was incomplete, but happened to work. Setuptools now warns about such a condition, which added noise in the build output.

`package_data` now includes the files downloaded/generated by scripts in ci/*.py.

`test_suite` was unnnecessary because pytest is used for testing. It has been deprecated since by setuptools 41.5.0 (Oct 2019), and its presence emitted a warning that added noise to build output.

This change should
- allow wasmtime-py wheels to be built, published, and installed.
- allow building of an sdist (source tar.gz distribution)

It should not
- ~~alter versioning behaviour (specifically .devNNN tagging)~~
- alter wheel contents

Risks
- `-C=--build-option=--plat-name -C=--build-option=...` is ugly and possibly fragile. A future release of build or setuptools may break it.
- No version constraints are placed on setuptools in pyproject.toml.

I think this ~~is the minimum viable change that~~ fixes #286. Other improvements can be made seperately (e.g. migrating more metadata from setup.py to pyproject.toml).

As of 2025-06-17 it als fixes #231, superceding #285.

See
- https://build.pypa.io/en/stable/
- https://packaging.python.org/en/latest/guides/writing-pyproject-toml/
- https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-test-suite
- https://peps.python.org/pep-0517
- https://github.com/pypa/pip/issues/11859
- https://github.com/pypa/setuptools/issues/2491
- https://github.com/pypa/setuptools/discussions/3882

Edit: reverted the packages change, due to a CI error.
Edit 2025-06-17: Updated scope and noted changes following incorporation of changes in #285.


moreati:
Bother, I'll take another look when I can - probably Monday.

alexcrichton:
Seems reasonable to me to have! I'll be honest that I don't really understand the calculus of all the various tools/configuration in play but I'm happy to merge green PRs when you feel they're ready

moreati:
So far, I think this is the approach I'll recommend - keep setuptools, use it as the backend to [build](https://build.pypa.io/en/stable/). I'll write up a pros/cons summary once I've ironed out the last kinks.

I explored switching from setuptools -> [hatchling](https://hatch.pypa.io/latest/config/build/#build-system) in [hatch-crimes branch](https://github.com/bytecodealliance/wasmtime-py/compare/main...moreati:wasmtime-py:hatch-crimes). Hatch has some version handling niceties. However I couldn't get it to output platform-specific wheels.

alexcrichton (approved):
All looks reasonable to me!

FWIW I think it's probably best nowadays to drop the test.pypi.org bits entirely, it seems like that was a bit too aspirational on my part

moreati:
Updated the description to note expanded scope. I think it's ready, thanks for sticking with me.

alexcrichton:
Thank you again for your help on this and your work!

alexcrichton:
Ah I set things up historically to publish all commits to test.pypi.org to make sure things work, and then it additionally publishes to pypi.org on tags. Ideally it would still publish to test.pypi.org but only when manually requested (e.g. by changing the version number) but I'm not sure how to publish-or-do-nothing depending on the version

PR #301, opened 2025-10-25
Add Nox configuration, to run tests locally across Python versions

moreati:
Currently it's awkward to run the testsuite locally across all supported Python versions, so version specific errors are only found in Github CI jobs.

This change allows the wasmtime-py test suite to be run across all supported CPython versions 3.9 - 3.14 at time of writing. It uses [Nox](https://nox.thea.codes/en/stable/index.html) a Python CLI tool for running tests or other build/project management tasks. It's invoked as
- `nox` or `python3 -m nox` - run every session defined in noxfile.py
- `nox --list` - list sessions defined
- `nox -s <session>` - Run a particular session
- `nox -p <python,...>` - Run sessions the use the given python version(s)

If desired checks such as MyPy or style can also be included.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -20,6 +20,7 @@ recursive-exclude wasmtime/bindgen/generated *.py *.wasm
 include .flake8
 include CONTRIBUTING.md
 include mypy.ini
+include noxfile.py
 include pytest.ini
 include VERSION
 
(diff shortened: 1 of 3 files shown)


moreati:
This PR is offered in the hope it will be useful. No hard feelings if you prefer not to take it, for any reason or no reason.

If you wish a similar capability, but less Python-centric, then [Just](https://just.systems/man/en/) might fit the bill. I used Nox because it's a hammer I know.

PR #332, opened 2026-04-03
Add wasi:http support

chaynabors:
Depends on bytecodealliance/wasmtime#12950 which adds wasmtime_component_linker_add_wasi_http and wasmtime_context_set_wasi_http to the C API.

Adds linker.add_wasi_http() and store.set_wasi_http() following the same pattern as add_wasip2 and set_wasi. _bindings.py was regenerated with cbindgen.py against the updated C headers.

### DIFF
--- a/tests/component/test_linker.py
+++ b/tests/component/test_linker.py
@@ -1,6 +1,6 @@
 import unittest
 
-from wasmtime import Engine, WasmtimeError, Module, Store
+from wasmtime import Engine, WasiConfig, WasmtimeError, Module, Store
 from wasmtime.component import Linker, Component, LinkerInstance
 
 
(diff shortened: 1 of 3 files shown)


alexcrichton (commented):
Thanks! This'll unfortunately have to wait until Wasmtime 44 is released since there's no easy way to pull in the dev binaries here for the time being. Mind adding some tests though which mostly just call the methods and ensure everything's wired up correctly? No need for a full end-to-end test as that'd be a pretty gnarly `*.wat`.

chaynabors:
Added tests and ran against my local build. Happy to wait since team has access to a release in my fork that I'm pointing at.

alexcrichton (approved):

PR #338, opened 2026-05-26
Update to Wasmtime 45.0.0

nekowinston:
Heya, I was running this on NixOS, and needed the `-std=c99` flag to avoid https://github.com/eliben/pycparser/issues/579 and other issues, like `bool` not being parsed as an acceptable type at https://github.com/bytecodealliance/wasmtime/blob/ea9d1fedfdeca401c570f09b9e08db328dc3d04d/crates/c-api/include/wasm.h#L199 for some reason.


alexcrichton (approved):
Thanks!

Issue #254, opened 2024-10-21
Make bindings threadsafe


alexcrichton:
Currently the bindings in this repository are not thread safe. This means it's possible to write Python code that's susceptible to race conditions and memory corruption. This is due to the fact that there are no locks in use by this library and the C API does not internally uphold thread-safety invariants, that's left to embedders.

There's some more discussion of this [here](https://bytecodealliance.zulipchat.com/#narrow/channel/217126-wasmtime/topic/.E2.9C.94.20Thread.20safety.20of.20wasmtime.20.28python.29.3F/near/477225037) and [here](https://github.com/bytecodealliance/wasmtime-py/issues/217#issuecomment-2035373713)

PR #342, opened 2026-08-24
Prevent `Slab` data corruption

barisa-cloudsmith:
### The bug

We've noticed the bug first when running main thread with wasmtime-py, where GC was run on threads that weren't using wasmtime-pyt ( boto upload); but same issue can happen on a single threaded code as well.
GC was running _del in a boto thread, and causing data corruption.

`Slab` threads its free list through the same array as its values — a free slot holds the index of the next free slot, a live slot holds the value — so the invariant spans two statements of `deallocate`, and there is no ownership check:

```python
def deallocate(self, idx: int) -> None:
    self.list[idx] = self.next
    self.next = idx
```

Any free that arrives out of turn leaves `self.next` pointing at a slot that is still live. The next `allocate` hands that slot out a second time — two owners on one handle, silently — and the one after indexes the list with the value it found there.

```python
from wasmtime._slab import Slab

slab: Slab[tuple] = Slab()
for i in range(3):
    slab.allocate(('func-%d' % i,))
victim = slab.allocate(('victim',))

slab.deallocate(victim)
slab.deallocate(victim)      # freed twice -- no ownership check

print('allocate ->', slab.allocate(('a',)))
print('allocate ->', slab.allocate(('b',)), '   <-- same handle, two owners')
print('allocate ->', slab.allocate(('c',)))
```

```
allocate -> 3
allocate -> 3    <-- same handle, two owners
  ...
  File "wasmtime/_slab.py", line 22, in allocate
    self.next = cast(int, self.list[idx])
                          ~~~~~~~~~^^^^^
TypeError: list indices must be integers or slices, not tuple
```

Once poisoned the slab stays poisoned: `self.next` is a tuple, so every subsequent allocation raises.

### The fix

Drop the free list rather than guard it. Handles come from `itertools.count(1)` and values live in a dict, so there is no invariant spanning two operations left to break:

- freeing a handle twice, or freeing one the caller never owned, is an idempotent `pop`
- handles are never reused, so a freed handle cannot come back as somebody else's
- handles are never `0`, so a null env cannot resolve to a live one
- `get` on a freed handle raises `KeyError` instead of returning a free-list integer

### Notes

- **No lock, deliberately.** #254 is clear that thread-safety invariants are the embedder's job, and I'm not trying to relitigate that here — this half is a single-threaded bug. There's also a concrete reason not to simply lock the existing methods: `deallocate` drops the last reference to its value, and that `__del__` can free another handle *inside* the outer call, so a plain `threading.Lock` around `deallocate` deadlocks. `test_deallocating_a_handle_can_free_another` pins that down. Under the GIL the individual dict and counter operations here are atomic; on free-threaded builds `itertools.count` isn't, so say the word if you'd like a lock. It wouldn't need to wrap these methods, though: scoping a plain `Lock` to the `next()` call alone is enough, and it keeps the lock off `deallocate`, where holding one across a value's `__del__` is what makes the reentrancy above a deadlock.
- `Slab` still has no `__len__` or `__bool__` on purpose — `_wasi.py` and `component/_linker.py` use the slab's truthiness as an interpreter-shutdown guard (`if CUSTOM_OUTPUTS:`), and giving it a length would make an empty slab skip the deallocate.
- Handles now grow monotonically instead of being recycled. They're passed as `void *`, so the ceiling is 2^64 allocations.
- Left out to keep this small: the `idx or 0` dances and `_value.py`'s `+1`/`-1` offset are all redundant now, and `Slab.get`'s `KeyError` could carry a better message. Happy to fold any of that in.
- Happy to split the count-from-1 change into its own commit or PR if you'd rather keep this to the free list alone.

### Testing

New `tests/test_slab.py`. Against the current implementation 3 of the 5 fail and `test_double_free_leaves_live_handles_alone` errors with the `TypeError` above; all 5 pass with this change.

Full suite on this branch: 184 passed, 2 failed — both failures are the pre-existing mypy error in `tests/test_module.py:11` (`bytearray` vs `str | bytes`), which reproduces unchanged on `main`.


alexcrichton:
Please make sure you review the Bytecode Alliance's [AI tool usage policy](https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md).

Otherwise deallocating the same index on a slab is a bug itself, not something the slab should be trying to fix. Could you detail more what the original bug here is?

barisa-cloudsmith:
> Please make sure you review the Bytecode Alliance's [AI tool usage policy](https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md).
> 
> Otherwise deallocating the same index on a slab is a bug itself, not something the slab should be trying to fix. Could you detail more what the original bug here is?

Thnx Alex.
Both points are valid: I didn't follow the AI tool usage policy. Also, the bug seems to be on the client usage.
I can't reproduce the bug on our version of python, using a single thread.

I'll close this PR for now, since the PR only helps with multiple threads usecase.

